PR #33, opened 2014-01-26
Moves garbage disposal to main thread and ensures it via SafeHandles.

myrup:
See http://en.sfml-dev.org/forums/index.php?topic=12248.30

I considered moving the disposal to Window.Clear(), but this necessarily isn't called. Like Window.DispatchEvents() is.

Merged ObjectBase and ObjectBaseSafeHandle since last pull request.


zsbzsb:
I am closing this, the issue has already been shown to be related to either stuff happening in multiple threads or not explicitly calling ```Dispose()``` when finished with objects. All this PR does is move the automatic GC to the main thread so it doesn't leak contexts.

PR #96, opened 2015-02-24
Added support for font info

zsbzsb:
Tested and working perfectly.

The only way to marshal a string inside a struct is to pass it as an IntPtr then use Marshal.PtrToStringAnsi(...) to convert it to a string. This is why I have two FontInfo structs in the PR.

http://stackoverflow.com/questions/22803083/methods-type-signature-is-not-pinvoke-compatible

### DIFF
--- a/src/Graphics/Font.cs
+++ b/src/Graphics/Font.cs
@@ -138,6 +138,22 @@ namespace SFML
 
             ////////////////////////////////////////////////////////////
             /// <summary>
+            /// Get the font information
+            /// </summary>
+            /// <returns>A structure that holds the font information</returns>
+            ////////////////////////////////////////////////////////////
+            public Info GetInfo()
+            {
+                InfoMarshalData data = sfFont_getInfo(CPointer);
+                Info info = new Info();
+
+                info.Family = Marshal.PtrToStringAnsi(data.Family);
+
+                return info;
+            }
+
+            ////////////////////////////////////////////////////////////
+            /// <summary>
             /// Provide a string describing the object
             /// </summary>
             /// <returns>String description of the object</returns>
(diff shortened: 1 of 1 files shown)


LaurentGomila:
As far as I remember, I never needed such a hack to unmarshal ANSI strings. It works automatically for arguments and return values, so I'm pretty sure there's a way to make it work also in structures.

And the new type should be Font.Info, not FontInfo.

* the same remarks apply to your other PR *

zsbzsb:
It spits out a runtime error any time I tried to use the string directly in the struct. I have tried every different resource I can find and most state to do it like I did (with IntPtr). The only source that says different (https://limbioliong.wordpress.com/2011/07/09/passing-managed-structures-with-strings-to-unmanaged-code-part-1/) I can't make work.

Not to mention this wouldn't be the only time we used an internal struct for marshaling data (https://github.com/LaurentGomila/SFML.Net/blob/master/src/Graphics/RenderStates.cs#L115).

http://stackoverflow.com/questions/1223690/pinvoke-error-when-marshalling-struct-with-a-string-in-it
http://stackoverflow.com/questions/1215335/marshalling-a-struct-containing-c-strings?lq=1
http://stackoverflow.com/questions/799076/returning-a-stdstring-from-a-c-dll-to-a-c-sharp-program-invalid-address-s
http://discuss.fogcreek.com/dotnetquestions/default.asp?cmd=show&ixPost=1108
https://groups.google.com/forum/#!msg/microsoft.public.dotnet.framework/vF0qxqjc0LA/bR_DkLHzB5gJ

zsbzsb:
PR has been updated.

LaurentGomila:
The name ```MarshalData``` worked for ```RenderStates``` because it was the marshal structure of the whole enclosing structure. But here it matches Font.Info, not Font itself. In other words, if one day we add another struct to the Font class, we won't know how to name the corresponding marshal data. So it should be something like ```Font.InfoMarshalData```.

zsbzsb:
I'm fine with doing that as long as you agree we need an internal structure for marshaling data. :smile:

LaurentGomila:
If you say

> String handling inside structs is a lot more complicated that as parameter/return types

Then I trust you. And since you found a similar thing in ```RenderStates```, then I also trust me :+1:.

LaurentGomila:
Now go fix #97 :stuck_out_tongue_winking_eye:

zsbzsb:
Already doing that  :smiley:

LaurentGomila:
I'm wondering... shouldn't this be a property?

zsbzsb:
Could be, but then we need to rename the Info type. Because you can't have a property by the same name as a type that is part of the class.

LaurentGomila:
What do you think, as a C# developer?

zsbzsb:
I think we should leave it as is. I don't see any outstanding reason to change it.

LaurentGomila:
Ok.

PR #97, opened 2015-02-25
Added support for joystick identification

zsbzsb:
Tested with XBox 360 controller and was able to get all the information. Also note that this is the same as #96 with regards to having an internal struct for marshaling the data.

### DIFF
--- a/src/Window/Joystick.cs
+++ b/src/Window/Joystick.cs
@@ -135,6 +135,56 @@ namespace SFML
                 sfJoystick_update();
             }
 
+            ////////////////////////////////////////////////////////////
+            /// <summary>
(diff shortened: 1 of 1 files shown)


LaurentGomila:
Same remarks as in #96.

zsbzsb:
And done :)

LaurentGomila:
VendorID -> VendorId
ProductID -> ProductId

In case you wonder, I checked and the names are ok in CSFML.

zsbzsb:
Well in C# we are capitalizing stuff. camelCase -> CamelCase  but this was a kinda unique case. :P

Updated... :)

PR #99, opened 2015-02-25
Updated SoundRecorder to support multiple input devices

zsbzsb:
And if you decide to mention that sfSoundRecorder_getDevice returns an IntPtr instead of a string that is required. Otherwise the interop will free our static string causing a crash on next call. And yes it has been fully tested.

http://elegantcode.com/2010/11/17/the-black-art-of-pinvoke-and-marshaling-in-net/
>Doing some more research on marshaling strings, I found out that in case the native function does its own cleanup, an IntPtr must be returned by the C# prototype declaration instead of a String.

### DIFF
--- a/src/Audio/SoundRecorder.cs
+++ b/src/Audio/SoundRecorder.cs
@@ -36,9 +36,9 @@ namespace SFML
             /// Please note that only one capture can happen at the same time.
             /// </summary>
             ////////////////////////////////////////////////////////////
-            public void Start()
+            public bool Start()
             {
-                Start(44100);
+                return Start(44100);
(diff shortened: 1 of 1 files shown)


LaurentGomila:
> And if you decide to mention that sfSoundRecorder_getDevice returns an IntPtr instead of a string that is required.

I don't care since it remains an internal detail.

But shouldn't GetDevice/SetDevice be a Device property instead?

zsbzsb:
setDevice returns a bool indicating if it was successful.

LaurentGomila:
Ah, right.

PR #105, opened 2015-03-08
Fixes Equals stack overflow

Meduax:
Previously a stack overflow would occur due to the two objects calling their respective object.Equals overrides over and over again. Now they call the IEquatable version as intended.

### DIFF
--- a/src/Graphics/BlendMode.cs
+++ b/src/Graphics/BlendMode.cs
@@ -157,7 +157,7 @@ namespace SFML
             ////////////////////////////////////////////////////////////
             public override bool Equals(object obj)
             {
-                return (obj is BlendMode) && obj.Equals(this);
+                return (obj is BlendMode) && Equals((BlendMode)obj);
             }
 
             ///////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


zsbzsb:
Good catch, just one thing, you don't need ```this.Equals(...)```. ```Equals(...)``` works just fine :wink:

src/Graphics/BlendMode.cs
```diff
@@ -157,7 +157,7 @@ public BlendMode(Factor ColorSourceFactor, Factor ColorDestinationFactor, Equati
             ////////////////////////////////////////////////////////////
             public override bool Equals(object obj)
             {
-                return (obj is BlendMode) && obj.Equals(this);
+                return (obj is BlendMode) && this.Equals((BlendMode)obj);
```
zsbzsb: return (obj is BlendMode) && Equals((BlendMode)obj);

Meduax:
Yeah, I know. this.Equals just conveys the intent more clearly in my opinion. I can change it though.

Meduax:
Changed.

zsbzsb:
Well @LaurentGomila would have the final say, but I suggested that because no where in the rest of the library do we use the ```this.Function(...)``` style.

Anyways, looks good now :+1:

LaurentGomila:
> this.Equals just conveys the intent more clearly in my opinion

The intent that must be highlighted is that we call the same function but with a typed argument, and the explicit cast does it very well. I don't see what "this" would make clearer ;)

Thanks for the contribution.

PR #107, opened 2015-03-23
Added AnyCPU build platform

dmirmilshteyn:
I added a build configuration option to allow compiling the .NET libraries under AnyCPU. They can be platform agnostic, but the CSFML variants for the platform need to be present in the search path during runtime. Eliminates some annoying warnings about platform mismatch when referencing the .NET libraries from an application compiled to run under AnyCPU.


zsbzsb:
>Eliminates some annoying warnings about platform mismatch when referencing the .NET libraries from an application compiled to run under AnyCPU.

The warnings aren't the problem. Projects still compile and execute. The trouble is that when projects run with x86 they won't work with x64 CSFML dlls and vice versa (targeting x64 won't work with x86 CSFML dlls). You will get a bad format exception at run time.

When you target 'AnyCPU' you let the JIT compiler determine what target you are going to use (x86 or x64). When that happens if the CSFML dlls don't match you also get a bad format exception. So anyone that decides to target 'AnyCPU' with SFML.NET and their own project will have issues if they ship x86 CSFML dlls on a x64 OS.

This is why I always tell people when using SFML.NET to either target x86 or x64 and not to use AnyCPU.

dmirmilshteyn:
I know, in most cases, it would be ideal to build for the platform being targeted. However, in my case, I need to build and distribute the application to run under any platform, and then add the search path to the CSFML dlls at runtime/through app.config.

This definitely shouldn't be the default build configuration, but it can be nice to have that option if someone wants to manage the native dependencies themselves.

zsbzsb:
>I need to build and distribute the application to run under any platform, and then add the search path to the CSFML dlls at runtime/through app.config.

Any chance you could elaborate on how this will work?

>This definitely shouldn't be the default build configuration

I agree we should try to support as many users as possible. But the way I see this is that it is an extreme edge case that will fail for just about any user.

zsbzsb:
Closing this for now, feel free to reopen if you can better explain how this will be used.

PR #109, opened 2015-03-28
Fixed SoundRecorder.DefaultDevice crashing

### LINKED ISSUE #108: SoundRecorder.DefaultDevice crashes program when used
Any attempts to use SoundRecorder.DefaultDevice (even as a watch while debugging) instantly crash the simple console program I'm working on. SoundRecorder.IsAvailable returns true, and I can use the strings in SoundRecorder.AvailableDevices without crashing the program. Output after the crash doesn't provide anything useful, but I'd be willing to compile a debug version of the SFML.NET libraries and post more info if no one can replicate this.

Currently using the most recent 64-bit release of SFML.NET in VS Express 2013 on a computer running Win7.

zsbzsb:
Fixes #108

### DIFF
--- a/src/Audio/SoundRecorder.cs
+++ b/src/Audio/SoundRecorder.cs
@@ -206,7 +206,7 @@ namespace SFML
             {
                 get
                 {
-                    return sfSoundRecorder_getDefaultDevice();
+                    return Marshal.PtrToStringAnsi(sfSoundRecorder_getDefaultDevice());
                 }
             }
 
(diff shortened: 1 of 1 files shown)


LaurentGomila:
Looks good, but can you explain what happened?

zsbzsb:
I will need to test this later to confirm, however I am pretty sure what happened. When I first implemented it I was testing with MSVC compiled libs and I think I just got lucky. If you remember, I had to implement the ```sfSoundRecorder_getDevice(...)``` function to return a pointer or else it would crash on the next call (the same way this was crashing on the first call). That is because if you define the return type in managed code to ```string``` (from ```const char*```) the managed language tries to then free the memory after converting it to a managed string.

PR #111, opened 2015-03-30
Bumped version number to 2.3 and updated readme.txt

zsbzsb:
So when it is time all we need to do is merge :smile:


zsbzsb:
As everything applicable from [CSFML #70](https://github.com/SFML/CSFML/issues/70) is completed it is time to merge this.

PR #113, opened 2015-04-16
Removed unneeded operations in Color class

dejangeci:
No need to & the bits, (byte)uint will discard MSB anyway.

### DIFF
--- a/src/Graphics/Color.cs
+++ b/src/Graphics/Color.cs
@@ -51,10 +51,13 @@ namespace SFML
             ////////////////////////////////////////////////////////////
             public Color(uint color)
             {
-                R = (byte)((color & 0xff000000) >> 24);
-                G = (byte)((color & 0x00ff0000) >> 16);
-                B = (byte)((color & 0x0000ff00) >> 8);
-                A = (byte)((color & 0x000000ff) >> 0);
+                unchecked
+                {
+                    R = (byte)(color >> 24);
+                    G = (byte)(color >> 16);
+                    B = (byte)(color >> 8);
+                    A = (byte)color;
+                }
             }
 
             ////////////////////////////////////////////////////////////
@@ -93,7 +96,7 @@ namespace SFML
                        " B(" + B + ")" +
                        " A(" + A + ")";
             }
-            
+
             ////////////////////////////////////////////////////////////
             /// <summary>
             /// Compare color and object and checks if they are equal
@@ -105,7 +108,7 @@ namespace SFML
             {
                 return (obj is Color) && Equals((Color)obj);
             }
-            
+
             ///////////////////////////////////////////////////////////
             /// <summary>
             /// Compare two colors and checks if they are equal
@@ -120,7 +123,7 @@ namespace SFML
                        (B == other.B) &&
                        (A == other.A);
             }
-            
+
             ////////////////////////////////////////////////////////////
             /// <summary>
             /// Provide a integer describing the object
@@ -129,10 +132,7 @@ namespace SFML
             ////////////////////////////////////////////////////////////
             public override int GetHashCode()
             {
-                return (int)(R << 24) |
-                       (int)(G << 16) |
-                       (int)(B << 8)  |
-                       (int)A;
+                return (R << 24) | (G << 16) | (B << 8) | A;
             }
 
             ////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


LaurentGomila:
Even if the result is the same, sometimes a litte more verbosity doesn't hurt, and makes things clear (ie. you don't have to read the spec of the language or the compiler to know what happens).

And you added colors that are not part of SFML for... what? To make sure that your PR will be rejected? :stuck_out_tongue:

dejangeci:
The original code was not syntactic sugar - those are actual operations that get executed on the CPU, and this commit cuts few that are redundant. It might not seem much, but Color might be instantiated many times in a multimedia app. Anyway, if someone inexperienced looks at the code `A = (byte)((color & 0x000000ff) >> 0);` it won't be any clearer than `A = (byte)color;`.

As for the colors, I find it a nice addition. I can generate the code for C++ SFML and create a pull request there, if that is what is required for this PR to go through.

LaurentGomila:
Compilers are smart, don't underestimate them. They are perfectly capable of optimizing this kind of simple arithmetic operations if they are redundant.

> Anyway, if someone inexperienced looks at the code ```A = (byte)((color & 0x000000ff) >> 0);``` it won't be any clearer than ```A = (byte)color;```.

To me it is: the first one means that we extract the latest 2 LSB, whereas the second one means that we cast the color to a byte.

> As for the colors, I find it a nice addition. I can generate the code for C++ SFML and create a pull request there, if that is what is required for this PR to go through.

Nope, if something that simple was wanted we would already have implemented it.

dejangeci:
> Nope, if something that simple was wanted we would already have implemented it.

Point taken - commit removed. Might be suitable for an extension library.

> To me it is: the first one means that we extract the latest 2 LSB, whereas the second one means that we cast the color to a byte.

And to me it is: the first one means that we zero-out all except last two bits, and then we *cast the color to a byte*, whereas the second one means that we *cast the color to a byte*. This is, of course, just for discussions sake, because both will get your head scratching when taken out of context. When you check out the full constructor, the whole thing seems obvious even if you don't know how bit-shifting works.

> Compilers are smart, don't underestimate them. They are perfectly capable of optimizing this kind of simple arithmetic operations if they are redundant.

I didn't - I ran tests with both methods in release mode, and found my code to be consistently ~12% faster (Code here: https://gist.github.com/dejangeci/93189dfa08a4b54bf831).

I saw the new Color ctor, saw the bit-shifting which got me thinking, and had some spare time to explore it and grow my own knowledge. I don't really mind the original code, it's perfectly fine.

zsbzsb:
I don't have anything really against changing it, just remember this is the way it is implemented in the C++ (and of course the C) versions. 

https://github.com/LaurentGomila/SFML/blob/master/src/SFML/Graphics/Color.cpp#L72

LaurentGomila:
Why ```unckecked``` by the way? And what about the ```(int)``` cast that you removed from ```GetHashCode```?

dejangeci:
> Why unckecked by the way?

Arithmetic overflow/underflow is a project-wide (.csproj) setting that is disabled by default (VS: Project -> Properties -> Build -> Advanced -> Check for arithmetic...). If some day someone decides to enable it, code will start to throw overflow exceptions. With `checked/unchecked` you can be explicit when to turn on and off this behavior regardless of project settings. 

```cs
byte b1 = unchecked((byte)uint.MaxValue); // Always fine
byte b2 = (byte)uint.MaxValue // Exception if arithmetic checks enabled in project
```

Personally, I always keep the checks enabled in debug builds to detect potential wraparound errors, but disable it in release for speed. The difference is only in the generated IL code - it calls a different opcode.

> And what about the (int) cast that you removed from GetHashCode?

Because the involved operands produce Int32 by default (http://stackoverflow.com/questions/9210373/why-do-shift-operations-always-result-in-a-signed-int-when-operand-is-32-bits) so there is no need to put 4 casts on four lines. Just a visual change, really.

zsbzsb:
I'm fine with the ```GetHashCode``` changes, but when it comes down to the changes in the ctor I really don't care either way, so @LaurentGomila what is the final say on this?

LaurentGomila:
Why not...

zsbzsb:
>Why not...

:+1: :steam_locomotive:

PR #120, opened 2015-07-19
Fixed project files for Mono 4.0

zsbzsb:
As reported [here](http://en.sfml-dev.org/forums/index.php?topic=18606.0) and the fix is based on a commit from [meebey/db4o-net](https://github.com/meebey/db4o-net). Please wait for a confirmation that this fixes the issue before merging. I tested and VS2008 is still fine with the files after the change.

https://github.com/meebey/db4o-net/issues/1
https://github.com/meebey/db4o-net/commit/c81dbd1590cef5229ed39adc45d52cc5f5f481aa


LaurentGomila:
Can you explain what the problems means, and why this fixes it?

zsbzsb:
I can't really explain it, googling for ```mono 4 could not import $(MSBuildBinPath)\Microsoft.CSharp.targets``` doesn't bring up much other than the issue I linked above. The only way to explain it is maybe Mono broke something with their importer, as I don't have the latest Mono installed to test it I am waiting to see if the OP on the forums responds.

zsbzsb:
Since it has been confirmed as fixing the issue I am going to merge this (and nothing has been said against it).

PR #122, opened 2015-08-23
Updated Glyph bounds to Int.

Zarickan:
Updated the bounds of Glyph to IntRect instead of FloatRect, as it in the C++ version is an IntRect and not a FloatRect.
See: http://www.sfml-dev.org/documentation/2.0/classsf_1_1Glyph.php#afe4cd37e5839955d7dd008e178d41f0c


zsbzsb:
>as it in the C++ version is an IntRect and not a FloatRect

In SFML 2.1 it is a IntRect, but in 2.2+ it is a FloatRect. Your link to the documentation is for 2.1.

PR #124, opened 2015-09-13
Added SolutionInfo as central versioning point

dejangeci:
Removed extra stuff from AssemblyInfos and added SolutionInfo as central versioning point so there is no need to update each project separately


zsbzsb:
So if I understand correctly this allows us to bump the version number from just one file for all projects? And since we are at it, why not move all common attributes (such as product name and copyright) into the same file?

dejangeci:
Yes, correct. I messed up with commits and squashing, so I'll make another pr

zsbzsb:
You don't need to make another PR, after you fix the branch you can force push and everything will be corrected.

dejangeci:
Took too much fiddling with git to get it right. Created a new pr

PR #125, opened 2015-09-21
Added SolutionInfo as central versioning point

dejangeci:
Removed extra stuff from AssemblyInfos and added SolutionInfo as central versioning point so there is no need to update each project separately

### DIFF
--- a/src/Audio/Properties/AssemblyInfo.cs
+++ b/src/Audio/Properties/AssemblyInfo.cs
@@ -1,33 +1,7 @@
 ﻿using System.Reflection;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-// Les informations générales relatives à un assembly dépendent de 
-// l'ensemble d'attributs suivant. Changez les valeurs de ces attributs pour modifier les informations
-// associées à un assembly.
 [assembly: AssemblyTitle("sfml-audio")]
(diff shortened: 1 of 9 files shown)


src/Audio/Properties/AssemblyInfo.cs
```diff
@@ -1,33 +1,8 @@
...
-[assembly: AssemblyCompany("")]
-[assembly: AssemblyProduct("SFML")]
-[assembly: AssemblyCopyright("Copyright © Laurent Gomila")]
-[assembly: AssemblyTrademark("")]
-[assembly: AssemblyCulture("")]
 
-// L'affectation de la valeur false à ComVisible rend les types invisibles dans cet assembly 
-// aux composants COM. Si vous devez accéder à un type dans cet assembly à partir de 
-// COM, affectez la valeur true à l'attribut ComVisible sur ce type.
 [assembly: ComVisible(false)]
```
zsbzsb: Not a big issue, but while you are changing this, would you stuff mind to move the com visible attribute also and set it to true? (all others except audio are visible)

dejangeci:
Only reason why I didn't move it was because Audio had ComVisible(false). I figured @LaurentGomila  had a  specific reason for this, but it was probably auto-generated like that.

dejangeci:
Done.

zsbzsb:
Tested and works great, thanks!

Issue #127, opened 2015-10-25
Null arguments for Shader(Stream, Stream) constructor


Slko:
Shader(Stream, Stream) constructor doesn't allow any of the arguments to be null.

LaurentGomila:
Indeed, this should be fixed.

zsbzsb:
Superseded by #129

PR #129, opened 2015-10-27
Allow passing null streams to Shader

Slko:
See #127

### DIFF
--- a/src/Graphics/Shader.cs
+++ b/src/Graphics/Shader.cs
@@ -75,11 +75,29 @@ namespace SFML
             public Shader(Stream vertexShaderStream, Stream fragmentShaderStream) :
                 base(IntPtr.Zero)
             {
-                StreamAdaptor vertexAdaptor = new StreamAdaptor(vertexShaderStream);
-                StreamAdaptor fragmentAdaptor = new StreamAdaptor(fragmentShaderStream);
-                CPointer = sfShader_createFromStream(vertexAdaptor.InputStreamPtr, fragmentAdaptor.InputStreamPtr);
-                vertexAdaptor.Dispose();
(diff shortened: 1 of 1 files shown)


zsbzsb:
Why not just utilize a ternary operator and avoid duplicating code?

```csharp
StreamAdaptor vertexAdaptor = vertexShaderStream != null ? new StreamAdaptor(vertexShaderStream) : null;
StreamAdaptor fragmentAdaptor = fragmentShaderStream != null ? new StreamAdaptor(fragmentShaderStream) : null;
CPointer = sfShader_createFromStream(vertexAdaptor != null ? vertexAdaptor.InputStreamPtr : IntPtr.Zero,
                                     fragmentAdaptor != null ? fragmentAdaptor.InputStreamPtr : IntPtr.Zero);
if (vertexAdaptor != null) vertexAdaptor.Dispose();
if (fragmentAdaptor != null) fragmentAdaptor.Dispose();
```

Slko:
The first version of my code looked like yours, but don't you think it's a mess? The only clean way I found was that one (which was inspired by the source code of sfShader_createFromStream, actually).
Other than that, both are okay.

zsbzsb:
>The first version of my code looked like yours, but don't you think it's a mess?

I personally like to try to avoid code duplication and IMO this situation is perfect for ternary operators.

>The only clean way I found was that one (which was inspired by the source code of ```sfShader_createFromStream```, actually).

I guess since it is already coded this way in CSFML we can keep the same style for consistency in SFML.Net.

>Other than that, both are okay.

Since I am fine with the changes now as is @LaurentGomila do you want to review this before I merge it?

LaurentGomila:
It looks good. So *if it has properly been tested*, it can be merged.

PR #130, opened 2016-01-28
Fixed incorrect logic in Color multiplication operator.

nidaegwyn:
The multiplication operator was incorrect.
I made the change based on the SFML C++ implementation of the multiplication operator.


zsbzsb:
For reference here is the C++ implementation. The changes look good to me :+1: 

```
return Color(Uint8(int(left.r) * right.r / 255),
    Uint8(int(left.g) * right.g / 255),
    Uint8(int(left.b) * right.b / 255),
    Uint8(int(left.a) * right.a / 255));```

LaurentGomila:
Looks ok.

PR #137, opened 2016-10-30
Add missing Context.Settings property, ContextSettings.SRgbCapable field, and Keyboard.SetVirtualKeyboardVisible() function

dabbertorres:
Context.SetActive() now returns bool for success/failure.

Bumped SolutionInfo.cs version to 2.4.

Matching CSFML commits:
SFML/CSFML@57eda5f9234a58d45d82f74ca8e4644478cd2118
SFML/CSFML@52c7de16f0e1d2c5d97e46598161b76cff7c6f18

### DIFF
--- a/src/Window/Context.cs
+++ b/src/Window/Context.cs
@@ -37,20 +37,22 @@ namespace SFML.Window
         /// Activate or deactivate the context
         /// </summary>
         /// <param name="active">True to activate, false to deactivate</param>
-		/// <returns>true on success, false on failure</returns>
+        /// <returns>true on success, false on failure</returns>
         ////////////////////////////////////////////////////////////
         public bool SetActive(bool active)
         {
             return sfContext_setActive(myThis, active);
         }
 
-		/// <summary>
-		/// Get the settings of the context.
-		/// </summary>
-		public ContextSettings Settings
-		{
-			get { return sfContext_getSettings(myThis); }
-		}
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Get the settings of the context.
+        /// </summary>
+        ////////////////////////////////////////////////////////////
+        public ContextSettings Settings
+        {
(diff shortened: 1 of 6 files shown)


LaurentGomila:
Why not a `Settings` property?

dabbertorres:
No real reason.

My logic was that since the settings must be retrieved (through the CSFML call), rather than just being able to return, a function may be a good choice to show that an action is taken.

LaurentGomila:
Although I tend to agree that actions should be functions, *everything* must be retrieved through a CSFML call (yes, it's a binding) so I would not take this in account ;)

dabbertorres:
True. Now that you mention that, I'm not exactly sure what my thought process was.

I'll make the change within a few days.

dabbertorres:
Fixed.

LaurentGomila:
Identation is completely messed up, probably because you use real tabs instead of 4 spaces.

dabbertorres:
Yup, forgot to format. Done.

src/Window/Context.cs
```diff
@@ -37,10 +37,19 @@ public Context()
...
+        public bool SetActive(bool active)
         {
-            sfContext_setActive(myThis, active);
+            return sfContext_setActive(myThis, active);
+        }
+
+        /// <summary>
+        /// Get the settings of the context.
+        /// </summary>
+        public ContextSettings Settings
```
LaurentGomila: Documentation blocks should be surrounded by ////////////////////////////////////////////////////////////

dabbertorres: Corrected

LaurentGomila:
I still see tabs.

dabbertorres:
I am not on top of things. Apologies.
Okay, no more tabs for certain this time. Also found and fixed some in another file.

LaurentGomila:
Looks good. Thanks.

PR #139, opened 2016-12-18
Dotnet core

Balnian:
As suggested by eXpl0it3r in:
http://en.sfml-dev.org/forums/index.php?topic=21248.0

This binding has been tested with Windows10.
/Exemple is broken and must be remade from ground up

Most commit in the branch are for fixing the project because i made it fit in the old library hierarchy 

Also some changes have been made because Dotnet Core doesn't support all the things the .Net Framework has.
See forum post linked earlier to see the most important changes made.

If there is any question don't hesitate to contact me :smile:


szotp:
I tried to open this in VS 2017, and it forces some migration, which doesn't work. Old version works fine without any migration needed.

Balnian:
Yeah... They changed the files (project.json -> projectName.csproj) it's the new Dotnet Core way of doing it ( the build system is now based on Msbuild and a xml file format) and since VS 2017 is not release yet i didn't provide a version for it. And the migration tool is just crap...

eXpl0it3r:
So will this work independently with VS 2015 and older, as well as .NET Core?

Make sure to squash the commits.

Balnian:
It will only work with VS2015 and newer, the .Net Core project doesn't exist prior to 2015.
For compatibility on Windows .NetFramework 4.5 is the minimum requirement.
[https://github.com/dotnet/standard/blob/master/docs/versions.md](https://github.com/dotnet/standard/blob/master/docs/versions.md)

Scellow:
I redone this from scratch here #141 with the samples, and port notes, with new csproj support

Balnian:
@Scellow 's PR #141  is more complete then mine, so it's a better candidate when all the fixes are made

dabbertorres:
It sounds all relevant people are in agreement - we can close this PR, @eXpl0it3r.

Kaev:
This MR is still open and can be closed as @dabbertorres mentioned.

PR #141, opened 2017-06-17
.NET Core 1.1 Port

Scellow:
Notes:

* Attribute ``SuppressUnmanagedCodeSecurity``  for native calls isn't present in .Net Core so i removed them, can't find any doc about it

* Serialization has been removed for .Net Core, it might come back for .Net Core 2.0 but it is not released yet, the only class that needed it was : ``LoadingFailedException`` 

* System.Runtime.ConstrainedExecution was removed too, there is no need for it in reality

* 🛑  There is no dllmap feature in .Net Core, that means we will have to build SFML.net for each platform individually, this might change for .Net Core 2.0

* I can't port opengl/window samples yet since OpenTK hasn't been ported to .Net Core yet, they plan to do it as soon as possible

* sfml-audio no longer need the sfml-window dependency

I'll update the notes if i find new stuff

Legend: 

* 🛑 = Needs to be fixed/implemented or to find a work-around to consider the port usable


Scellow:
Hmm i think i'll just port the examples to .NET Core too

examples/window/Window.cs
```diff
@@ -1,169 +0,0 @@
-using System;
```
dabbertorres: No point in removing examples that work on other platforms as mentioned previously.

src/Audio/Listener.cs
```diff
@@ -69,28 +69,28 @@ public static Vector3f UpVector
         }
         
         #region Imports
-        [DllImport("csfml-audio-2", CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
```
dabbertorres: Do you have any measurements on the performance difference of removing `SuppressUnmanagedCodeSecurity`?

SFML has always been for "no unnecessary optimizations", so since this was added in the first place, it's likely here for a reason.

If the performance difference is minimal though, I don't see why removing it is a problem!

src/Audio/Properties/AssemblyInfo.cs
```diff
@@ -1,7 +0,0 @@
-﻿using System.Reflection;
```
dabbertorres: Does .NET Core not have AssemblyInfo?

src/System/LoadingFailedException.cs
```diff
@@ -1,14 +1,12 @@
...
 
-namespace SFML
+namespace SFML.System
 {
     ////////////////////////////////////////////////////////////
     /// <summary>
     /// Exception thrown by SFML whenever loading a resource fails
     /// </summary>
     ////////////////////////////////////////////////////////////
-    [Serializable]
```
dabbertorres: I don't think I have any issues with this change (Serializable doesn't exist in .NET Core, correct?)

src/System/LoadingFailedException.cs
```diff
@@ -1,14 +1,12 @@
 using System;
-using System.Runtime.Serialization;
 
-namespace SFML
+namespace SFML.System
```
dabbertorres: Unnecessary namespace change

src/System/StreamAdaptor.cs
```diff
@@ -2,7 +2,7 @@
 using System.Runtime.InteropServices;
 using System.IO;
 
-namespace SFML.Window
```
dabbertorres: Unnecessary, but, makes sense. Fine with me.

src/Window/Context.cs
```diff
@@ -10,7 +9,7 @@ namespace SFML.Window
     /// This class defines a .NET interface to an SFML OpenGL Context
     /// </summary>
     //////////////////////////////////////////////////////////////////
-    public class Context : CriticalFinalizerObject
```
dabbertorres: CriticalFinalizerObject is necessary last I checked to ensure cleanup of OpenGL contexts.

src/Audio/Listener.cs
Balnian: it's not a performance optimisation, it just dosn't exist in .Net Core

src/Window/Context.cs
Balnian: If i remember correctly, from my .Net Core Port CriticalFinalizerObject dosn't exist in .Net Core, but it should have been replaced at least by IDisposable see my "Dotnet Core" PR ;)

src/Audio/Listener.cs
dabbertorres: https://msdn.microsoft.com/en-us/library/system.security.suppressunmanagedcodesecurityattribute(v=vs.110).aspx

It makes .NET not perform a stack walk when calling a native function.
Direct quote from the link: "The stack walk performed when calling unmanaged code is omitted at run time, resulting in substantial performance savings."

src/Window/Context.cs
dabbertorres: IDisposable != CriticalFinalizerObject

https://blogs.msdn.microsoft.com/cbrumme/2004/02/20/finalization/

src/Audio/Listener.cs
dabbertorres: Found a comment from a CoreCLR dev: https://github.com/dotnet/coreclr/issues/1605
It's essentially a "built-in" feature for .NET Core.

So yes, not needed in .NET Core, but would be needed for normal .NET.

Balnian:
> There is no dllmap feature in .Net Core, that means we will have to build SFML.net for each platform individually, this might change for .Net Core 2.0

if i remember correctly, DllImport do some platform matching like it will try for "csfml-window-2":

1. csfml-window-2.dll
2. libcsfml-window-2.so
3. ...<wathever Mac use>csfml-window-2.<wathever Mac use>

but it's been a long time since i read that and cannot find the article so take this with a pinch of salt ;)

src/Audio/Listener.cs
Balnian: maybe, but if you leave it in, it won't compile for .Net Core, that's why my old PR  doesn't have them too #139

Balnian: and sadly, when converting the project to .Net Core you will not have the same compatibility
see [this](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-platforms-support) so user will need to at least target .NET Framework 4.5

src/Audio/Properties/AssemblyInfo.cs
Balnian: yes it does see [here](https://docs.microsoft.com/en-us/dotnet/api/system.reflection?view=netcore-1.1)

src/Window/Context.cs
Balnian: When looking for a replacement for this in .Net Core, this was the best next thing available, because you still need to release the resources

examples/visualbasic/My Project/Application.Designer.vb
```diff
@@ -1,13 +0,0 @@
-﻿'------------------------------------------------------------------------------
```
Balnian: Vb is not supported for .Net Core 1.1  it has been added to .Net Core 2.0  stated [here](https://blogs.msdn.microsoft.com/dotnet/2017/05/10/announcing-net-core-2-0-preview-1/)

Rohansi:
I believe .NET Core does those [convention-based lookups](https://github.com/dotnet/coreclr/pull/1780) still. 

However these do not work for architecture differences. The NuGet packages support OS and architecture specific libraries though so a single package could be referenced to provide builds for all supported platforms.

src/Audio/Properties/AssemblyInfo.cs
dabbertorres: Thanks.

@Scellow can you revert the removals of AssemblyInfo.cs?

dabbertorres:
I appreciate the comments and clarifications @Balnian and @Rohansi!

At the moment, given VB support and CriticalFinalizerObject not arriving until .NET Core 2.0, I'm leaning towards postponing until then. Preview versions started last month it looks like, so the full release should be coming up relatively soon<sup>TM</sup> I think.

Of course, need to hear from @LaurentGomila and others from the SFML team (ie: @eXpl0it3r)!

Balnian:
I made a forum post so we can discuss about how we should bring SFML to .Net Core:
[here](https://en.sfml-dev.org/forums/index.php?topic=22150.0)

examples/shader/Program.cs
```diff
@@ -13,10 +13,7 @@ protected Effect(string name)
             myName = name;
         }
 
-        public string Name
-        {
-            get { return myName; }
-        }
+        public string Name => myName;
```
Scellow: I see no reasons to stick with older version of C#, if port to .NET Core, better to stick with modern C#

src/System/LoadingFailedException.cs
Scellow: it was necessary to remove ``sfml-graphics`` dependency for ``sfml-audio``

src/System/LoadingFailedException.cs
Scellow: It doesn't, but i heard it is going to come back for 2.0 i'll have to check that again

examples/shader/Program.cs
dabbertorres: I can agree with that.
Need to hear from @LaurentGomila

src/System/LoadingFailedException.cs
dabbertorres: Ah, good choice then. Appreciate it!

src/System/LoadingFailedException.cs
dabbertorres: +1

examples/shader/Program.cs
LaurentGomila: > Need to hear from @LaurentGomila

What does it involve in terms of compatibility?

dabbertorres: The major benefit is native support for OSX and Linux (and wherever else .NET Core can be compiled), so Mono would no longer be needed on those platforms. Additionally, a smaller runtime dependency.
Additionally, .NET Core is much more detached from Visual Studio versions. Even on Windows, VS is not needed, only the Core SDK.

.NET Core 1.x ships with C# 6 support minimum (I don't know if .NET Core 2 will up it to 7 or not).

As for operating systems:
Windows 7+
OSX 10.11+
Linux (not really any change as far as I can tell)

Hopefully @Scellow / @Balnian / @Rohansi can comment further.

My thoughts: MS doesn't support (doesn't want to at least) XP and Vista anymore, does SFML still want to support XP and Vista? I don't know/use OSX enough to have an opinion for it.

If you haven't had a chance to read much else, it's looking like waiting for .NET Core 2.0 to be released (and updating the PR) before merging will be best ([summary](https://en.sfml-dev.org/forums/index.php?topic=22150)).

LaurentGomila: My question was more about backward compatibility with older (non-Core) .Net framework versions

Balnian: I've made a forum post regarding .Net Core upgrade [here](https://en.sfml-dev.org/forums/index.php?topic=22150.0)
But basicaly if we target .Net Core 1.1 (this PR) the minimum version of .Net Framework needed to use the the binding would be 4.5 or greater

LaurentGomila: > 4.5 or greater

Sounds acceptable.

dabbertorres: What's the Framework version needed for Core 2.0?

Balnian: it's 4.6.1 for Core 2.0 [Table for lookup](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-platforms-support)

dabbertorres: @LaurentGomila that work for you too? We were thinking of waiting for .NET Core 2.0 so VB.NET support isn't dropped, and also to wait for OpenTK to support 2.0.

LaurentGomila: > We were thinking of waiting for .NET Core 2.0

When will it be released?

dabbertorres: According to the [roadmap doc](https://github.com/dotnet/core/blob/master/roadmap.md), Q3, this year.

Kaev:
This can probably be closed aswell since we have the .NET Standard/Core 2.0 MR

Balnian:
I agree with @Kaev , we should close it.

MaximeMazet:
Hello
How i can download the SFML.net for the .NET Core ?

Thanks

dsoronda:
@MaximeMazet .. 
#150 should be merged soon so stay tuned
Until then, you can download code from [here](https://github.com/dsoronda/SFML.Net), build binaries and reference them in your projects.

dsoronda:
We can close this one now :)

PR #142, opened 2017-08-09
.Net Core/Standard 2.0 support

Balnian:
Based on #141 (Credits to @Scellow)

- Some Exemples are broken (missing OpenTK for 2.0)
- We will need to redo the visualbasic exemple because it use
WindowsForms or maybe it should target .Net Framework?
- We also need to change the way we map DLLs because the name will change on Linux/macos


Balnian:
Use PR #143  instead

PR #143, opened 2017-08-11
.Net Standard/Core 2.0 + DllImport Mapping

Balnian:
New PR with DllImport Mapping
Based on #141  (Credits to @Scellow)
- Some Exemples are broken (missing OpenTK for 2.0)
- the visualbasic exemple is back and also broken because of OpenTk... (why is there so much OpenTK in the exemples ???)

PR #150, opened 2018-02-22
Updated projects to .NET Core & included changes from PR #143

dsoronda:
Changes:

- Created new .NET Core projects and migrated code into them
- Updated nuget references
- Fixed project build by replacing OpenTK local references with [OpenTK.NETCore](https://www.nuget.org/packages/OpenTK.NETCore) nuget package
- Merged changes from PR #143 (includes some code refactoring, etc...)
- Added publish profiles for projects
- Removed OpenTK from `window` example, since we already have `opengl` example
- other stuff I couldn't remember but are in comments below...


dsoronda:
Can anyone code review this PR so that @LaurentGomila could merge this back?

How about this : put +1 on this comment if you approve code merge :)

dsoronda:
Guys, did anyone test this ?

Two months passed and no feedback. I really don't want to push unofficial [SFML.Netstandard](https://github.com/dsoronda/SFML.Netstandard) as alternative.

dsoronda:
Can we move on with this ?
Even Microsoft is ditching .NET Framework.

Is there anyone willing to test and confirm that this work ?

texus:
I can run the shader example (using dotnet restore/build/run commands in terminal) from both this branch and #143 on Manjaro linux without issues (with SFML and CSFML installed in /usr/local/lib).

Running the opengl example gave me a `System.PlatformNotSupportedException: Operation is not supported on this platform.` exception from OpenTK, but its already better than not having the examples work on any platform.

Maybe someone should summarize what should be tested. One of the main things that is holding me back with helping (other than the fact that I know nothing about dotnet core) is that it is unclear what still has to happen. What is the difference between this PR and #143? How should these branches be tested? How should SFML.Net be build (and used in a project) after these changes are merged?
I can test on linux but also on windows and macOS in virtualbox if needed, but I don't know enough about dotnet to know what to test exactly.

dsoronda:
OpenGL examples should not work (yet) since there are no OpenTK for Netstandard.
I found this one [OpenTK.NETCore](https://www.nuget.org/packages/OpenTK.NETCore) but still didn't have time to test it.

Differences between this PR and #143 are:

1. I created all new .Net core project files with template names like `[old_project_name]_core.csproj` using `dotnet new classlib` templates and copied all code from original branch.
2. Updated nuget references
2. Added manually all changes from #143
3. Code cleanup
4. Fixed Window example to use Draw instead OpenTK so it can run properly (we already have OpenGl example).
5. Tested all examples to work on Windows (except OpenGL)

To compile examples you should : 
- clone my [fork](https://github.com/dsoronda/SFML.Net)
- change directory to `build/vs2017/` and use `dotnet build SFML.net.sln`
- go to examples directory and copy native SFML libraries in examples build directory (bin/debug/netstandard...) so .net  apps can find them (or add SFML libraries to path)
- run examples with `dotnet run example.dll`.

I'm preparing Linux machine for testing but I'm little busy with my 3 month old baby.

Sadly, all those changes breaks original VS 2015 solution and project files but since Microsoft is dumping .Net Framework it doesn't matter so much.

texus:
I managed to use your branch to build and use SFML.Net in a test project on Windows, Linux and macOS. I only tested building and then running via `dotnet run`, I didn't try creating an executable with `dotnet publish` yet.

On windows the native libraries have to be copied to the build directory like you said, but this of course doesn't work on linux and mac as it doesn't search for them in the working directory. When the native libraries are installed globally then running `dotnet run` works out-of-the-box. Alternatively the LD_LIBRARY_PATH (linux) or DYLD_LIBRARY_PATH and DYLD_FRAMEWORK_PATH (mac) environment variables can be set to where the libraries can be found.

I used the sln in build/SfmlCore because the sln from the vs2017 folder couldn't find reference assemblies for .NETFramework. To run the examples, `dotnet run example.dll` didn't work because there are multiple project files. I had to run `dotnet run example.dll -p shader_core.csproj`.

I've noticed that the versions of the SFML dlls are still 1.0.0.0, this should probably be changed by adding `<Version>2.4.2</Version>` in the PropertyGroup section in the *_core.csproj files. I don't know about the other csproj files, why do they even exists? Is it just for compatibility?

I haven't tested the opengl example on windows yet, but it does run on mac. So basically it crashes on windows, doesn't support linux and runs just fine on mac.

texus:
I tried using OpenGL.Net instead of OpenTK and it seemed to work, but it only worked on windows (not on linux or mac) and I don't even think the nuget supported .net frameworks, only .net core projects. So that didn't look like a good solution either.

So I tried to find a workaround for the crash in OpenTK. Since the nuget that works with .netcore is very outdated, it is likely that the bug has already been fixed in the meantime. It unfortunately still doesn't look like OpenTK is going to target .net standard anywhere soon (its planned for 4.0 while they haven't released 3.0 yet).

After trying some random things hoping to work around the bug, I succeeded in getting it to run. Not only does it work without crashing on windows now, it also runs on linux (and still runs on mac). The workaround should probably be removed if OpenTK ever gets official .net standard support, but in the meantime, the following patch seems to fix the opengl example:
```diff
diff --git a/examples/opengl/OpenGL.cs b/examples/opengl/OpenGL.cs
index 3029210..71bec85 100644
--- a/examples/opengl/OpenGL.cs
+++ b/examples/opengl/OpenGL.cs
@@ -16,6 +16,8 @@ namespace opengl
         /// </summary>
         static void Main()
         {
+            var tmp = new OpenTK.GameWindow();
+
             // Request a 24-bits depth buffer when creating the window
             var contextSettings = new ContextSettings();
             contextSettings.DepthBits = 24;
@@ -25,8 +27,8 @@ namespace opengl
             window.SetVerticalSyncEnabled(true);
 
             // Initialize OpenTK
-            Toolkit.Init();
-            GraphicsContext context = new GraphicsContext(new ContextHandle(IntPtr.Zero), null);
+            //Toolkit.Init();
+            //GraphicsContext context = new GraphicsContext(new ContextHandle(IntPtr.Zero), null);
 
             // Setup event handlers
             window.Closed     += new EventHandler(OnClosed);
```

dsoronda:
Whoa, great job @texus.

I updated project version to 2.4.2 and fixed this bug in OpenGL example.

`To run the examples, dotnet run example.dll didn't work because there are multiple project files. I had to run dotnet run example.dll -p shader_core.csproj.`

Yeah, sorry for misunderstanding, I was already in `bin\debug...\` directory when I was typing those instructions (I had to manually copy CSFML dlls insite) so I (wrongly) assume that others will know that to.

Once again, you really did great job @texus.

I hope that @LaurentGomila will merge this request soon and publish new nuget packages.

texus:
I would reduce the TargetFramework of the bouncing ball example to 2.0, there is no need for it to require 2.1 (which was only released 2 days ago).

Maybe a comment should also be added in the opengl example at the OpenTK.GameWindow() line stating that it is only a workaround to create a functioning opengl context with that OpenTK version. It might otherwise be confusing why the line exists.

dsoronda:
Changes in latest commit:

Removed old vs2010 solution
Reduced TargetFramework of the bouncing ball example to Core 2.0
Added some comment to OpenGL example regarding OpenTK

eXpl0it3r:
I'm interested in moving forward with SFML.Net and thought I'll start here, as you've been quite active.

The general "issue" with this PR is, that the title says one thing, but the PR does a lot more, which unfortunately also isn't documented in the description. This leads to either having to ask a ton of questions or having to blindly accept the changes. The latter is not really what we do and I guess for the first one, nobody has invested the time so far.

Would be great if you could answer the following questions and update the PR description, so it's immediately clear what this PR does.

- What do the batch files in `build/SfmlCore` do?
- Why are these batch files needed?
- Why was Bouncing Ball added as example?
- What are the `*.config` for?
- Why do we need all the project files with `_core`?
- Why are Assembly Title commented out?

Going through the change set quickly, I noticed some general remarks:

- `.gitignore` shouldn't be added
- We rather stick to our current formatting with spaces, instead of changing everything to tabs
- Since this should be an update for the solution, it really shouldn't contain any refactorings (using `var`, lambdafying functions, etc.)

As a general tip, it's recommended to keep changes focused on one topic only and to provide reasoning for the changes. That way it's a lot easier for everyone to understand what's going on and why it was done. 🙂

Balnian:
> What do the batch files in build/SfmlCore do?

I don't think we should have them, because build.bat is only a wrapper over `dotnet build`, they should learn to use the normal dotnet tools and not a wrapper. 
Also for publish.bat/sh only appears to be a helper to build and copy the audio and other SFML.Net "projects" into the publish/build directory, This should be handle as project references in the csproj.


> Why do we need all the project files with _core?

If I remember correctly, it was to keep a single repo for .Net Framework and .Net Core.
Do we really want to keep this like this or do we want to create a brand new repo/org for .Net Core?

And I think all the other things that @eXpl0it3r raised are valid, to simplify the merge, we might want to remove those extra and add them after we did the merge for .Net Core

Btw, I should now be notified when there's activity here.

Also, I don't know for all of you but do you think it would be reasonable to try to merge this before September? Would be nice to maybe put ourself a challenge/deadline because we're so near the finish line!

dsoronda:
> The general "issue" with this PR is, that the title says one thing, but the PR does a lot more, which unfortunately also isn't documented in the description. This leads to either having to ask a ton of questions or having to blindly accept the changes. The latter is not really what we do and I guess for the first one, nobody has invested the time so far.

This PR started as simple project migration to .NET Core but I continued to work on it. I'm open to suggestion how to change title.

@eXpl0it3r,  I'm confused about `documented in the description` part. I can add file `changes_PR150.md` or something similar to list changes in PR if it helps, or did you meant something else ?

> > What do the batch files in build/SfmlCore do?
> 
> I don't think we should have them, because build.bat is only a wrapper over dotnet build, they should learn to use the normal dotnet tools and not a wrapper.
> Also for publish.bat/sh only appears to be a helper to build and copy the audio and other SFML.Net "projects" into the publish/build directory, This should be handle as project references in the csproj.

Bat/sh files are removed.

> Why was Bouncing Ball added as example?

Removed

> What are the *.config for?

They were added when I referenced new OpenTK Core lib.

> >Why do we need all the project files with _core?
 
> If I remember correctly, it was to keep a single repo for .Net Framework and .Net Core.
> Do we really want to keep this like this or do we want to create a brand new repo/org for .Net Core?

I created clean *_Core project and kept old project for references so I could easily merge changes without messing up old project files. Sadly, .NET Core add breaking changes so I kept old project files in case I missed something.
IMHO, we should abandon old .NET Framework and keep .NET Standard (*_core) projects and have only one repo/org. I can rename old projects into *_obsolete or something.

> Why are Assembly Title commented out?

Because, VS 2017 throws this error when I un-comment it.

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0579	Duplicate 'System.Reflection.AssemblyTitleAttribute' attribute	sfml-system_core	w:\github\SFML.Net\src\System\obj\Debug\netstandard2.0\sfml-system_core.AssemblyInfo.cs	19	Active

> What's FolderProfile.pubxml for?

I think those are publish profiles for Core app. More info [here](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/visual-studio-publish-profiles?view=aspnetcore-2.1&tabs=aspnetcore2x).

> When was the addition of LoadingFailedException.cs discussed?

This has nothing to do with this PR. That file was already in master or PR #143 

>.gitignore shouldn't be added

I don't agree with you on this one. It's there with reason. Like, some peoples (like me) uses VS Code and `.vscode/` line should be in `.gitignore`. Not to mention /bin etc...

> We rather stick to our current formatting with spaces, instead of changing everything to tabs

I don't agree with you on this one. It's tabs vs spaces flame war and I don't want to go there.

>Since this should be an update for the solution, it really shouldn't contain any refactorings (using var, lambdafying functions, etc.)

I don't agree with you on this one. You are forgetting that this is not just simple project/solution update since I **included all changes from #143** into this PR. So all tabs/var/lambda stuff are mostly from there.

This refactorings are simple aesthetics and doesn't change functionality in any way of project. Same is valid for tabs vs spaces argument.

I spend **a lot of hours** fixing this PR and merging stuff from #143 and in testing, and will not waste any more time in fixing tabs or changing back var or lambda functions back to getter methods.

For tabs vs spaces arguments, I'll recommend that in next PR we add `.editorconfig` & `omnisharp.json` (for VS Code) files that will set coding standards for future merges.

dsoronda:
Guys, @Balnian @texus @LaurentGomila 

Microsoft is putting End of life on .NET Framework. Can we move on with this PR ???

I really don't want to fork this project but this is taking too much, 7 months for PR approval?

dsoronda:
Only reason I used %_core naming is to keep old Project files as references in case I missed something in transition. I could rename _core projects to proper names and suffix old projects file with _old. Is that ok ?

texus:
Personally I think we should just have .NET Standard libraries with examples for .NET Core. I wouldn't even keep the old project files.

I think its fine for the additions to .gitignore to be kept. I understand that eXpl0it3r wants to keep it minimal, but at the same time I also prefer not to have clutter in the `git status` output. It doesn't do much harm to have a big .gitignore file, it might even prevent people from accidentally pushing too many files in the future.

Similarly, I don't mind the refactoring (e.g. usage of `var`). Sure, ideally it shouldn't have been part of this PR, but I wouldn't expect it to be undone just to make this PR slightly cleaner and then have the changes made again for a new PR. It's been over a year now since #143 was created, if SFML.Net would have been actively maintained then I would consider this a valid issue, but given the situation I think we should be glad that something is being done instead of nitpicking on details. As long as the code is still clean and it doesn't cause issues, I think it isn't a problem to have a few lines too many changed as part of such a huge update.

I do think we should be consistent with spaces/tabs and use spaces everywhere. Even if the tabs came from the other PR, they should still be fixed. Although I don't remember the exact commands I used, I converted all tabs to spaces in my project some time ago so I know it doesn't take that that long. If @dsoronda isn't willing to make these changes then I can always do that for him.

I think the answer to "Why are Assembly Title commented out?" isn't good enough. If there is an error then it should be known why. I looked into it and it seems AssemblyInfo is partially generated by the contents of the csproj file (this can be easily disabled but then even more info has to be added to AssemblyInfo.cs). I propose we just delete the AssemblyInfo.cs (together with the old projects). The AssemblyDescription can be added to the csproj file (it's just called "Description" there), the Guid doesn't seem necessary. I also noticed that the version in the dll is still 1.0.0, so we might as well add the version now we are at it. So in the first `<PropertyGroup>` in the *_core.csproj file something like this should be added:
```
<Description>Graphics module of the SFML library</Description>
<Version>2.4.2</Version>
```

In the end it is up to @eXpl0it3r (or anyone else in the SFML team that cares) to decide when the PR is good enough. But I think it should be made clear what the remaining blocking points are, so that they can be solved so that this PR can finally be merged.

dsoronda:
`.gitignore` is generated with new Visual studio solution with support for Git. Is just standard setup.
I converted all tabs to spaces (it is possible that i missed some).

I added `.editorconfig` and `omnisharp.json` to enforce spaces and other C# related stuff, so it will help with future PR.

> I think the answer to "Why are Assembly Title commented out?"
Fixed, there were duplicates in `AssemblyInfo.cs` for Description filed.

Now, few notes:
- **I don't want to waste any more of my time on converting tabs or any stuff that is ecstatic in nature vs functionality that will help to improve this project.**
- This  is my last change  on this PR. I'll welcome any PR to my fork (SFML.Net) that will go to this PR.
- I created new fork [SFML.Netstandard](https://github.com/dsoronda/SFML.Netstandard) and it will contain all changes from my fork of SFML.Net. I'll give time till 2018-10-15 to official SFML team to accept this merge, after that I'll try to publish my SFML.Netstandard fork to nuget.

I really wish to use this project but maybe it's time to move on.

Best of luck to team

@Balnian  @texus @LaurentGomila @eXpl0it3r @zsbzsb

ashleydavies:
Just wanted to add another opinion that .NET Standard is a great place to have SFML.NET development in future. I’ve already moved my project over to @dsoronda’s fork (thanks for porting it!) to be able to build it cross-platform without Mono, which was really straightforward and conivent.

If the SFML(.NET) team is too busy to help work on the changes - which is totally understandable! - then perhaps it’s worth endorsing or separately publishing the fork in the meantime? This repository could be kept for anyone who wants to stay with the .NET Framework version for any reason?

I’m happy to also help contribute to this PR if it looks like this can get landed, as per @dsoronda’s comment above.

Either way, thanks to all of the SFML developers for making such a great library! I’d love to see a future with an official .NET Standard port, however that is achieved. And apologies if I should’ve posted this on the forum or somewhere else instead.

Kaev:
I can only agree with the posts above me. This should totally be merged.

LaurentGomila:
If there's a general agreement on this PR then I'll merge it.

Let's wait a few days, to make sure we've heard all voices.

eXpl0it3r:
I think it's okay to merge now and clean up stuff afterwards. 🙂

Issue #156, opened 2018-10-19
Font byte[] constructor


FRex:
There is strong suspicion we use a pin, take address of a memory block and then unpin the block, so if we read from it much later it'll fail if the GC has moved it.

The offending code is in src/Graphics/Font.cs:
```C#
public Font(byte[] bytes) : base(IntPtr.Zero)
{
    GCHandle pin = GCHandle.Alloc(bytes, GCHandleType.Pinned);
    try
    {
        CPointer = sfFont_createFromMemory(pin.AddrOfPinnedObject(), Convert.ToUInt64(bytes.Length));
    }
    finally
    {
        pin.Free();
    }
    if (CPointer == IntPtr.Zero)
    {
        throw new LoadingFailedException("font");
    }
}
```
Docs of this functionality: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandletype?view=netframework-4.7.2

Came up in this thread (and supposedly using a MemoryStream instead works as fix without requiring changes to C/C++ side): https://en.sfml-dev.org/forums/index.php?topic=24650

DemoXinMC:
This issue can be closed, as it was corrected in 2.5.0

PR #157, opened 2018-11-10
Update SFML.Net to 2.5.0

DemoXinMC:
This brings SFML.Net up to SFML 2.5. Additionally addresses #156 by switching the Font(byte[]) constructor to convert the raw byte[] to a MemoryStream and calling Font(Stream) instead

### DIFF
--- a/src/Graphics/sfml-graphics.csproj
+++ b/src/Graphics/sfml-graphics.csproj
@@ -26,7 +26,7 @@
 		<IsOSX>False</IsOSX>
 		<IsAndroid>False</IsAndroid>
 		<IsLinux>False</IsLinux>
-		<Version>2.5.1</Version>
+		<Version>2.5.0</Version>
 		<Description>Graphics module of the SFML library</Description>
 	</PropertyGroup>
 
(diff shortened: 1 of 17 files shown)


src/Graphics/sfml-graphics.csproj
```diff
@@ -26,7 +26,7 @@
 		<IsOSX>False</IsOSX>
 		<IsAndroid>False</IsAndroid>
 		<IsLinux>False</IsLinux>
-		<Version>2.4.2</Version>
+		<Version>2.5.1</Version>
```
eXpl0it3r: As this will be the first release it should be SFML.Net 2.5.0, that it uses SFML 2.5.1 underneath is irrelevant, meaning you could just as well use SFML 2.5.0 and it won't make a difference.

src/Window/sfml-window.csproj
```diff
@@ -6,7 +6,6 @@
 		<AssemblyName>sfml-window</AssemblyName>
 		<Configurations>Debug;Release;_WINDOWS_;_LINUX_;_OSX_</Configurations>
     <Description>Window module of the SFML library</Description>
-    <Version>2.4.2</Version>
```
eXpl0it3r: No replacement needed?

DemoXinMC:
Requested changes made in 3rd commit of the PR

src/Window/sfml-window.csproj
DemoXinMC: Not sure why this isn't marking as Outdated

Issue #166, opened 2019-09-16
ContextSettings Signature Mismatch (MarshalDirectiveException)


rfuge:
I just upgraded to SFML 2.5 and found attempting to read Settings property on a RenderWindow throws a MarshalDirectiveException with the message _Method's type signature is not PInvoke compatible_

Here's my code to reproduce.
```csharp
class Program
{
    static void Main(string[] args)
    {
        var running = true;
        using (var window = new RenderWindow(VideoMode.DesktopMode, "Test"))
	{
	    window.Closed += (sender, eventargs) => running = false;
            var settings = window.Settings;

            while (running)
	    {
	        window.WaitAndDispatchEvents();
		window.Clear();
		window.Display();
	     }
	}		
    }
}
```
I had a dive into the source and think this is because in CSFML the SRgbCapable member of ContextSettings is an sfBool which is Int32 and .NET is expecting a boolean.

rfuge:
Just to test my conclusion I swapped out SRgbCapable in ContextSettings to be an Int instead of a bool and that resolves the issue.

@eXpl0it3r Where would be the incentive to fix this? Would ContextSettings struct be changed in CSFML or in SFML.Net?

eXpl0it3r:
What's the underlying type in C++? If it's a mismatch between SFML and CSFML, fix it in CSFML. If it's a mismatch between CSFML and SFML.Net, fix it in SFML.Net.

rfuge:
in both C++ and C# its bool and in C its sfBool (Int)

I see sfBool is used everywhere in place of bool in CSFML so I'd be unsure if changing it for this one instance would be desired.

eXpl0it3r:
Ah I see. I'll have to investigate how to resolve this. It seems like for functions we simply use `bool` in the import statements, not sure if a similar thing can be done for ContextSettings struct.

eXpl0it3r:
Should probably be able to do something with the `[MarshalAs(...)]` attribute on the [SRgbCapable property](https://github.com/SFML/SFML.Net/blob/master/src/SFML.Window/ContextSettings.cs#L116).

rfuge:
So I was expecting `[MarshalAs(UnmanagedType.Bool)]` described as a [4-byte boolean](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=netframework-4.8) to resolve this but I still get the the same exception for some reason.

DemoXinMC:
[...]

PR #167, opened 2019-09-17
Fix #162 and #166

### LINKED ISSUE #162: Mapping error for sfRenderTexture_createWithSettings (CSFML 2.5.0)
The RenderTexture should use the new ctor with the ContextSettings:
 
```
[DllImport( CSFML.Graphics, CallingConvention = CallingConvention.Cdecl ), SuppressUnmanagedCodeSecurity]
static extern IntPtr sfRenderTexture_createWithSettings( uint Width, uint Height, ContextSettings Settings );
```
Unfortunately, sfRenderTexture_createWithSettings is not exposed in CSFML 2.5.0 native.
To easily exhibit the error, redefine the default ctor implentation of RenderTexture like it SHOULD be (since the one with bool depthBuffer is deprecated):
```
        public RenderTexture(uint width, uint height) :
            this(width, height, default(ContextSettings))
        {
        }
```
And runs the shader.csproj sample. Kaboom!!

And I don't understand why since it is actually part of the C code base:
https://github.com/SFML/CSFML/blob/282621e4995b5a22c4c5981e8f2cad98807d3554/include/SFML/Graphics/RenderTexture.h#L68

DemoXinMC:
This should address issues with reading ContextSettings from RenderWindow, as well as allow creation of RenderTextures by only dimensions without deprecation warnings.

Also provided XML documentation for a number of "undocumented" methods/enums/classes


src/SFML.Window/ContextSettings.cs
```diff
@@ -112,7 +113,10 @@ public override string ToString()
         /// <summary>The attribute flags to create the context with</summary>
         public Attribute AttributeFlags;
 
+        /// <summary>Internal Representation</summary>
+        private int _SRgbCapable;
+
         /// <summary>Whether the context framebuffer is sRGB capable</summary>
-        public bool SRgbCapable;
+        public bool sRGBCapable { get => _SRgbCapable == 0; }
```
rfuge: Accidental value flip? shouldn't it be `_SRgbCapable == 1;`?

DemoXinMC: You, you're right.  I stared at different boolean types for far too long.

PR #183, opened 2020-02-16
Update Time.cs

DemoXinMC:
Added conversions to and from .NET's TimeSpan struct
Significantly reduced P/Invoke calls, which now only happen during construction and conversion

This should yield performance gains in any place where Time is being used. Additionally, this is a bit of futureproofing for SFML3, as Time is likely to be removed, and the TimeSpan struct has existed in .NET since Framework 1.0

### DIFF
--- a/src/SFML.System/Time.cs
+++ b/src/SFML.System/Time.cs
@@ -48,24 +48,47 @@ namespace SFML.System
 
         ////////////////////////////////////////////////////////////
         /// <summary>
+        /// Construct a Time value from a TimeSpan
+        /// </summary>
+        /// <param name="timeSpan">A TimeSpan representing the amount of time to represent</param>
+        /// <returns>Time value constructed from an existing TimeSpan</returns>
+        ////////////////////////////////////////////////////////////
+        public static Time FromTimeSpan(TimeSpan timeSpan) => sfMicroseconds(timeSpan.Ticks * 1000 / TimeSpan.TicksPerMillisecond);
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
         /// Returns the time value as a number of seconds
         /// </summary>
         ////////////////////////////////////////////////////////////
-        public float AsSeconds() => sfTime_asSeconds(this);
+        public float AsSeconds() => microseconds / 1000000f;
 
         ////////////////////////////////////////////////////////////
         /// <summary>
         /// Returns the time value as a number of milliseconds
         /// </summary>
         ////////////////////////////////////////////////////////////
-        public int AsMilliseconds() => sfTime_asMilliseconds(this);
+        public int AsMilliseconds() => (int)(microseconds / 1000);
 
         ////////////////////////////////////////////////////////////
         /// <summary>
         /// Returns the time value as a number of microseconds
         /// </summary>
         ////////////////////////////////////////////////////////////
-        public long AsMicroseconds() => sfTime_asMicroseconds(this);
+        public long AsMicroseconds() => microseconds;
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Returns the time value as a TimeSpan
+        /// </summary>
+        ////////////////////////////////////////////////////////////
+        public TimeSpan ToTimeSpan() => TimeSpan.FromTicks(microseconds * (TimeSpan.TicksPerMillisecond / 1000));
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Implicit conversion from TimeSpan to SFML.System.Time, allowing intuitive use
+        /// </summary>
+        ////////////////////////////////////////////////////////////
+        public static implicit operator Time(TimeSpan timeSpan) => FromTimeSpan(timeSpan);
 
         ////////////////////////////////////////////////////////////
         /// <summary>


src/SFML.System/Time.cs
```diff
@@ -46,26 +46,49 @@ public struct Time : IEquatable<Time>
...
         public static Time FromMicroseconds(long microseconds) => sfMicroseconds(microseconds);
 
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Construct a Time value from a TimeSpan
+        /// </summary>
+        /// <param name="timeSpan">A TimeSpan representing the amount of time to represent</param>
+        /// <returns>Time value constructed from an existing TimeSpan</returns>
+        ////////////////////////////////////////////////////////////
+        public static Time FromTimeSpan(TimeSpan timeSpan) => sfMicroseconds(timeSpan.Ticks / 10);
```
eXpl0it3r: I might be wrong, but I don't think, you can just convert between ticks and time like this, or can you?
[Ticks](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.ticks?view=netstandard-2.0#System_TimeSpan_Ticks) doesn't seem to have a specific unit and I guess there's a reason why [`TimeSpan.TicksPerMillisecond`](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.tickspermillisecond?view=netstandard-2.0) exists.

src/SFML.System/Time.cs
```diff
@@ -46,26 +46,49 @@ public struct Time : IEquatable<Time>
...
         ////////////////////////////////////////////////////////////
-        public long AsMicroseconds() => sfTime_asMicroseconds(this);
+        public long AsMicroseconds() => microseconds;
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Returns the time value as a TimeSpan
+        /// </summary>
+        ////////////////////////////////////////////////////////////
+        public TimeSpan ToTimeSpan() => TimeSpan.FromTicks(microseconds * 10);
```
eXpl0it3r: See comment above

src/SFML.System/Time.cs
```diff
@@ -46,26 +46,49 @@ public struct Time : IEquatable<Time>
...
         ////////////////////////////////////////////////////////////
-        public float AsSeconds() => sfTime_asSeconds(this);
+        public float AsSeconds() => microseconds / 1000000f;
 
         ////////////////////////////////////////////////////////////
         /// <summary>
         /// Returns the time value as a number of milliseconds
         /// </summary>
         ////////////////////////////////////////////////////////////
-        public int AsMilliseconds() => sfTime_asMilliseconds(this);
```
eXpl0it3r: If we don't use sfTime_asMilliseconds and sfTime_asSeconds, should we still define the CSFML DLL reference at the bottom of the file?

src/SFML.System/Time.cs
DemoXinMC: .NET Standard has `TimeSpan.TicksPerMillisecond` as a const `10000`, making 1 microsecond = 10 ticks.

src/SFML.System/Time.cs
DemoXinMC: It's not hurting anything for them to be there, and they do exist in the CSFML lib, which is why I left them, even unused. I can remove them for clarity sake if we think that's best.

src/SFML.System/Time.cs
eXpl0it3r: Shouldn't you still use the const instead of writing your own hard coded values? And does the standard require this, or is that just one implementation of the standard?

DemoXinMC: It is the Standard, the const is for clarity, but will actually require a division operation to do the milliseconds to microseconds conversion, as there is not a `TicksPerMicrosecond` field.  The implementation I've done here is virtually identical to SFML's native one.

LaurentGomila: I'm pretty sure that the C# compiler can pre-compute the result of a constant divided by another constant, and not add extra runtime CPU cycles for it.

LaurentGomila:
> This should yield performance gains in any place where Time is being used

Is this something that you have measured? Rewriting parts of SFML instead of direct calls, has a maintenance cost and should only be done if really necessary.

DemoXinMC:
I specifically discovered the performance hit when using `SFML.System.Time`  as part of my animating sprite setup and discovered it to be the single largest bottle neck I have. P/Invoke calls add up very quickly, and considering that this is a very minor change for a class that isn't very volatile, I think this is a worthwhile bit of code copying.

`TimeSpan` exists within .NET, and serves an identical purpose, but this is the simplest way to allow the use of `TimeSpan` without rewriting existing signatures to cover a custom marshaler.

When SFML 3 becomes available, I plan to remove the use of `sfTime` in favor of `TimeSpan` and a custom marshaler.

src/SFML.System/Time.cs
DemoXinMC: I can switch this to use the `TicksPerMillisecond` constant, but it is currently identical to the native implementation, and also will not truncate accuracy as it is currently.

LaurentGomila: > it is currently identical to the native implementation

Which native implementation are you referring to?

> will not truncate accuracy as it is currently

Where is it truncating accuracy?

DemoXinMC: The native implementation in `sf::Time::` for converting microseconds to seconds/milliseconds.

The truncation of accuracy from using `TicksPerMillisecond` would result from microseconds between milliseconds being lost.

LaurentGomila: > The native implementation in sf::Time:: for converting microseconds to seconds/milliseconds.

Milli/micro/seconds are SI units and their conversion factors are well known... On the other hand, system "ticks" may be anything, that's why they provide such a constant to perform conversions.

> The truncation of accuracy from using TicksPerMillisecond would result from microseconds between milliseconds being lost.

I don't how using a named constant rather than a numeric constant would change anything.

DemoXinMC: Using milliseconds runs the number of ticks through a lower granularity.  Additionally, from [MSDN](https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=netframework-4.8)`Time values are measured in 100-nanosecond units called ticks.`

DemoXinMC: Upon further review, I can change this to use the constant.  I will update my PR shortly.

DemoXinMC: PR is updated and commits squashed.

eXpl0it3r:
While I see the cost for re-implementing code, I do think this one is justified, given that it's quite common to call the conversion functions a lot and you wouldn't expect P/Invoke calls for such a simple operation.

DemoXinMC:
Any further thoughts on this?

LaurentGomila:
```
timeSpan.Ticks / (TimeSpan.TicksPerMillisecond / 1000)
```

I know it's not the case here, but if `TimeSpan.TicksPerMillisecond` was not an exact multiple of 1000, this would be incorrect. A 100% safe (future-proof) version would simply be:

```
timeSpan.Ticks * 1000 / TimeSpan.TicksPerMillisecond
```

DemoXinMC:
Those 2 formulas yield the same result even with non-multiples of 1000

LaurentGomila:
In floating-point arithmetic yes, not with integer division -- unless C# has different rules than C++ for that.

DemoXinMC:
Fair enough.  Formula changed and PR updated.

Issue #192, opened 2020-08-07
RenderTexture constructor throws


DumpsterDoofus:
## Minimal Working Example

```
new RenderTexture(100, 100, new ContextSettings());
```

Running this throws the following:

```
System.EntryPointNotFoundException: Unable to find an entry point named 'sfRenderTexture_createWithSettings' in DLL 'csfml-graphics'.
   at SFML.Graphics.RenderTexture.sfRenderTexture_createWithSettings(UInt32 Width, UInt32 Height, ContextSettings Settings)
   at SFML.Graphics.RenderTexture..ctor(UInt32 width, UInt32 height, ContextSettings contextSettings)
   at Tiling.Program.Main() in E:\Software\My Software\GraphicsGarbage\Tiling\Program.cs:line 38
```

Happens when https://github.com/SFML/SFML.Net/blob/2.5/src/SFML.Graphics/RenderTexture.cs#L533 tries to call https://github.com/SFML/CSFML/blob/eb4a9cfe1def386132d45e7366926c731b74f15a/src/SFML/Graphics/RenderTexture.cpp#L57. 

More info:

- Package: `SFML.Graphics` 2.5.0
- Framework: `netcoreapp3.1`
- OS: Windows 10

## Why it breaks

https://github.com/SFML/CSFML/releases/tag/2.5 was released on Dec 8, 2018, but with an incorrect method signature (see https://github.com/crumblingstatue/CSFML/blob/eb4a9cfe1def386132d45e7366926c731b74f15a/include/SFML/Graphics/RenderTexture.h#L68).

On Mar 14, 2019 https://github.com/crumblingstatue/CSFML/commit/95ce89c3b7046e4103aebc5eb0efc721f100cdb4 fixed the method signature. However, there hasn't been a new release of CSFML, so SFML.Net is still on CSFML 2.5, which has the incorrect method signature.

Solution: Wait for CSFML 2.6, and update the version?

DangerRoss:
I too am waiting on when this will be available, mostly so I can set an AA level on a fullscreen RenderTexture.

The main alternative right now is to use the obsolete constructor for RenderTexture. But yes, I believe this is waiting on a new CSFML version to be released which I don't know when thats planned.

DangerRoss:
The fix for this issue was included in CSFML 2.5.1 https://github.com/SFML/CSFML/releases/tag/2.5.1

SFML.Net currently is still on 2.5.0 so this is likely to be resolved in the next release, when ever that might be.

[...]

PR #201, opened 2021-06-20
Update RenderTexture's ctor import

nulldg:
The RenderTexture constructor used in SFML.NET, `createWithSettings`, does not exist in CSFML 2.5 and causes exceptions when attempting to create a RenderTarget instance.
Unless I'm mistaken, this should be `create`.

### DIFF
--- a/src/SFML.Graphics/RenderTexture.cs
+++ b/src/SFML.Graphics/RenderTexture.cs
@@ -54,7 +54,7 @@ namespace SFML.Graphics
         /// <param name="contextSettings">A ContextSettings struct representing settings for the RenderTexture</param>
         ////////////////////////////////////////////////////////////
         public RenderTexture(uint width, uint height, ContextSettings contextSettings) :
-            base(sfRenderTexture_createWithSettings(width, height, ref contextSettings))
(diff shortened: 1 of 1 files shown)


DemoXinMC:
https://github.com/SFML/CSFML/blob/master/include/SFML/Graphics/RenderTexture.h#L68

createWithSettings absolutely DOES exist.

nulldg:
Using the CSFML 2.5 binaries downloaded from the SFML website, the current codebase results in this error
``
System.EntryPointNotFoundException: 'Unable to find an entry point named 'sfRenderTexture_createWithSettings' in DLL 'csfml-graphics'.'``
Was `createWithSettings` added in SFML 2.5.1 but the official build of CSFML doesn't yet support it?

DangerRoss:
This looks like its related to this issue raised awhile ago... https://github.com/SFML/SFML.Net/issues/192

Its a signature mismatch in CSFML which is fixed but not released IIRC

nulldg:
Oh, fair enough. My apologies. I'll have to find a way to build CSFML, and just hope the new version gets released soon.

PR #202, opened 2021-09-12
Add an allocation-free alternative to Image.Pixels

Slko:
Currently, there are two ways to access the pixels of an `Image`.

The first one is `Image.GetPixel()`, which does a call to a native function on each invocation, which is pretty bad for performance if you do it millions of times per second.

The other one is `Image.Pixels`  
It allocates a new pixel buffer on each getter access. Property accessors should never do anything heavy inside of them (according to common sense and [the official design guidelines](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100))), you can easily shoot yourself in the foot by writing a naïve code like this:  
```csharp
for (int y = 0; y < image.Size.Y; y++)
{
    for (int x = 0; x < image.Size.X; x++)
    {
        // Do something with image.Pixels[(y * image.Size.X + x) * 4]
    }
}
```  
Which _seems_ faster than calling `Image.GetPixel()`, but is actually `O(W * H)` allocations of `O(W * H)` bytes!

This pull request adds a new allocation-free and more performant alternative, which copies the internal pixel buffer to a user-provider byte array, and is a safer alternative to #138.

And I would also recommend you to deprecate `Image.Pixels` for the reasons stated above.

### DIFF
--- a/src/SFML.Graphics/Image.cs
+++ b/src/SFML.Graphics/Image.cs
@@ -299,6 +299,24 @@ namespace SFML.Graphics
 
         ////////////////////////////////////////////////////////////
         /// <summary>
+        /// Copy the array of pixels (RGBA 8 bits integers
+        /// components) to the provided buffer.
+        /// Buffer size must be at least Width x Height x 4
+        /// </summary>
+        /// <param name="buffer">Buffer to copy to</param>
+        ////////////////////////////////////////////////////////////
+        public void CopyPixels(byte[] buffer)
+        {
+            Vector2u size = Size;
+            if (buffer.Length < size.X * size.Y * 4)
+            {
+                throw new ArgumentException("Buffer must have a size of at least Width * Height * 4 bytes", nameof(buffer));
+            }
+            Marshal.Copy(sfImage_getPixelsPtr(CPointer), buffer, 0, (int)(size.X * size.Y * 4));
+        }
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
         /// Size of the image, in pixels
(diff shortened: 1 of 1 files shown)


DemoXinMC (approved):
Anything that we can use as a stop-gap until we transition to having access to Span<T> strikes me as a positive.

DemoXinMC:
Upon further review (yes, this PR is 2 years old), I don't like how this is implemented.  

Slko is correct that `Pixels` should not be allocating and then returning a `byte[]`, but fixing this would be a breaking change.  Ultimately, the `Pixels` shouldn't exist as a property. [CA1819: Properties should not return arrays](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/code-quality/ca1819-properties-should-not-return-arrays?view=vs-2015).  There are no redeeming qualities to `Pixels` continuing to exist other than avoiding breaking changes.

I would rather see this as `byte[] Image.GetPixels()` which would allocate the correct size buffer based on the image size, and `void Image.GetPixels(byte[])` that would fill the provided `byte[]` rather than allocating for itself.

There's also an argument to be made that these functions should be working with `Color[]` instead of `byte[]` and have separate methods (`.GetPixelBytes()` and `.GetPixelBytes(byte[])` or similar) for working with pixels on raw level.  `.GetPixel(int,int)` and `.SetPixel(int,int,Color)` already work with `Color`.

This can also all be simplified once we have access to `Span<T>` as we would avoid copying back and forth from managed to unmanaged memory.

PR #208, opened 2022-01-02
Fix #207

### LINKED ISSUE #207: Music.TimeSpan fields should be public
Is there any reason the Music.TimeSpan fields are private? It's not possible to set any meaningful loop points without having access to them.

```csharp
// Music.cs

/// <summary>
/// Structure defining a Time range. 
/// </summary>
[StructLayout(LayoutKind.Sequential)]
public struct TimeSpan
{
    Time offset;
    Time length;
}
```

DangerRoss:
Fix #207 
Makes Music.TimeSpan fields public

For maintainers;
I used [Rect.cs](https://github.com/SFML/SFML.Net/blob/master/src/SFML.Graphics/Rect.cs#L209) as a reference on how to make TimeSpan fields public. This means I've changed the field names to be capitalised and haven't used properties. As the fields were never accessible even internally to start with, the name change shouldn't cause a API breaking change.

### DIFF
--- a/src/SFML.Audio/Music.cs
+++ b/src/SFML.Audio/Music.cs
@@ -414,7 +414,7 @@ namespace SFML.Audio
         static extern TimeSpan sfMusic_getLoopPoints(IntPtr Music);
 
         [DllImport(CSFML.audio, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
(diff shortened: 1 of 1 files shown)


DangerRoss:
This now includes fix for pinvoke method signature that was found to have the incorrect return type during feature testing this PR

eXpl0it3r (approved):
Thanks a lot for following up on this and tracking down the bug! 🙂

PR #209, opened 2022-01-13
FIxes Missing Access Modifier Warnings

DangerRoss:
This PR is a refactor across SFML.Net that resolves all 457 warnings of missing access modifiers (IDE0040). My original motive to do this was due to missing access modifiers being part of bug #207. 

All 457 warnings are missing access modifiers for the pinvoke methods and so this PR adds `private` modifier to all pinvoke methods as thats the default when not specified.

Some diffs show entire source files being changed which is a result of some source files having mixed file endings and I've set it to whatever I found to be the common in the relevant project.

In terms of testing, I've done some basic feature tests such as shape rendering and playing music. As this is a sweeping change across the entire project, **I encourage some thorough code reviewing and understand if this PR gets put on hold for a future major version change.**

This wasn't just a find and replace effort and was an opportunity to also check over every PInvoke method and their CSFML counterparts, which has uncovered other issues of various severity that I'll raise separately after creating this PR.

### DIFF
--- a/src/SFML.Audio/Listener.cs
+++ b/src/SFML.Audio/Listener.cs
@@ -70,28 +70,28 @@ namespace SFML.Audio
 
         #region Imports
         [DllImport(CSFML.audio, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
-        static extern void sfListener_setGlobalVolume(float Volume);
+        private static extern void sfListener_setGlobalVolume(float Volume);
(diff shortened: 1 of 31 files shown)


eXpl0it3r (approved):
Looks good to me, thanks to the "hide white space" feature, it was quite easy to review 👍

eXpl0it3r:
Thanks again for this clean up 🧹

Issue #213, opened 2022-01-13
GLSL Bool Vector Types Maybe Mismatched With CSFML


DangerRoss:
The following PInvoke methods use the bool vector types and I believe won't actually work.

- sfShader_setBvec2Uniform
- sfShader_setBvec3Uniform
- sfShader_setBvec4Uniform

All Bvec types in C# use `bool` which is 1-byte and in CSFML they are `sfBool` which is 4-bytes. While this has never noted to be a problem with `bool` pinvoke parameters, it has been an issue in the past for data structures #166. I haven't used shaders in SFML.Net so I haven't confirmed this but i'm pretty confident its an issue.

Radix-Comet:
Just tested it, it works as expected.
![image](https://user-images.githubusercontent.com/104397149/171524404-8f4bafd3-f30f-44ef-8603-411baa1f0fd5.png)

Radix-Comet:
Just tested it, yes it is an issue. I'm working on a fix right now

eXpl0it3r:
Fixed with #217

PR #217, opened 2022-06-02
Updated GLSL.cs

Radix-Comet:
Mentioned in #213: CSFML specifies defines ``sfBool`` as ``int`` which is a 4-byte long integer. Until decided, I have marked the fields to be marshalled as ``Bool`` which seems to fix the issue. This will probably be to be fixed later. but it works for now on 64x OS's


eXpl0it3r (approved):
Looks good!

Have you tested this to work as expected?

PR #224, opened 2022-12-28
Update CSFML Dependency to 2.5.1

### LINKED ISSUE #223: Texture.Update(Texture, uint, uint) not found?
Tried to use the `Texture.Update` method to update from another texture, but got a C++ exception about missing entry point:

```
Exception thrown: 'System.EntryPointNotFoundException' in SFML.Graphics.dll
An unhandled exception of type 'System.EntryPointNotFoundException' occurred in SFML.Graphics.dll
Unable to find an entry point named 'sfTexture_updateFromTexture' in DLL 'csfml-graphics'.
```

https://github.com/SFML/SFML.Net/blob/f678eb2cf1eadf1a4f95cf1febd1da1d78ddcd7f/src/SFML.Graphics/Texture.cs#L230-L233

(Also looks like the overload without the x, y values and just taking the `Texture` directly is missing?)

image

Using NuGet package 2.5.0.

Code:

```cs
RenderTexture buff1 = new(800, 600);
Texture buff2 = new(800, 600);

// ... Draw stuff to buff1 ...

buff2.Update(buff1.Texture, 0, 0); // Fails
```

-----

I was creating a whole new Texture each time before (which works), but is eating memory and causing issues with Garbage Collection:

```cs
Sprite buff2Sprite = new();

buff2Sprite.Texture = new Texture(buff1.Texture);
```

Is there some other way I should be doing this operation?

hawkerm:
Fixes #223

Discovered CSFML 2.5.0 is missing a method export on the `Texture` class. 2.5.1 has the missing method.

This PR updates the `Version` property in the root props file which is used here to reference the same versioned CSFML package:

https://github.com/SFML/SFML.Net/blob/f678eb2cf1eadf1a4f95cf1febd1da1d78ddcd7f/SFML.Module.props#L20

Thanks!

### DIFF
--- a/SFML.NuGet.props
+++ b/SFML.NuGet.props
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
 
-    <Version>2.5.0</Version>
+    <Version>2.5.1</Version>
     <Authors>Laurent Gomila</Authors>
     <PackageTags>sfml sfml.net</PackageTags>
     <Copyright>Copyright © Laurent Gomila</Copyright>


eXpl0it3r:
I've implemented support directly for CSFML 2.5.2 and adjusted a few more things, so this PR has been superseded by #228

PR #233, opened 2023-10-29
Handle null pointers in Cursor constructors

### LINKED ISSUE #232: Using an unsupported CursorType causes a crash
The following code causes a segmentation fault on Debian 12. This is due to CSFML's `sfCursor_createFromSystem` returning a null pointer for unsupported cursors, and C#'s cursor taking that pointer and using it as-is.

```cs
using SFML.Window;

namespace SFMLRepro;

public static class Program
{
    public static void Main(string[] args)
    {
        using var window = new Window(new VideoMode(1280, 720), "SFML Cursor Crash");
        
        var cursorType = Cursor.CursorType.ArrowWait;
        
        using var cursor = new Cursor(cursorType);
        window.SetMouseCursor(cursor);
    }
}
```

I'd expect there to be an exception explaining the issue instead of the application hard crashing from this, since the C++ lib returns booleans to communicate success or failure of the operation. I think there's other places in the code that might have a similar issue (like the other constructor that calls `sfCursor_createFromPixels` and also takes the pointer as-is).

Marioalexsan:
Fixes #232 by throwing `LoadingFailedException` when a null pointer is returned from CSFML.

### DIFF
--- a/src/SFML.Window/Cursor.cs
+++ b/src/SFML.Window/Cursor.cs
@@ -122,9 +122,14 @@ namespace SFML.Window
         /// the operating system.
         /// </summary>
         /// <param name="type">System cursor type</param>
+        /// <exception cref="LoadingFailedException" />
         public Cursor(CursorType type)
             : base(sfCursor_createFromSystem(type))
         {
+            if (CPointer == IntPtr.Zero)
+            {
+                throw new LoadingFailedException("cursor");
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/SFML.Window/Cursor.cs
```diff
@@ -122,9 +122,14 @@ public enum CursorType
         /// the operating system.
         /// </summary>
         /// <param name="type">System cursor type</param>
+        /// <exception cref="LoadingFailedException" />
         public Cursor(CursorType type)
             : base(sfCursor_createFromSystem(type))
         {
+            if (CPointer == IntPtr.Zero)
+            {
+                throw new LoadingFailedException("cursor");
```
eXpl0it3r, line 131: The error message generated from this constructor could be bit misleading, but I think it's okay

eXpl0it3r (approved):

Issue #222, opened 2022-12-25
Add proper support for sRGB


eXpl0it3r:
Due to the way the CSFML API was implemented for creating/loading textures, there was no way to create textures in sRGB mode and the exposed setSrgb method was essentially useless.

With https://github.com/SFML/CSFML/pull/177 this was fixed, which then however leaves a gap in SFML.NET. The property for sRGB needs to be adapted and the new creation functions need to be called from SFML.NET.

PR #236, opened 2024-01-03
Update Texture for CSFML's sRGB Support

DemoXinMC:
Addresses #222
* Updated Texture's constructors with an optional parameter named "sRGB" which selects between the sRGB and non-sRGB versions of sfTexture_create*
* Allows Texture(byte[]) to specify a portion of the passed bytes to use for the new Texture, consistent with other existing constructors


eXpl0it3r:
Superseded by #252

Regardless, we appreciate the work! 🙂

Issue #212, opened 2022-01-13
Inconsistent Type Mapping Of Size_T Parameters


DangerRoss:
PInvoke methods and callbacks with `size_t` parameters or returns is inconsistently typed across SFML.Net.

Majority use `uint`
- sfVideoMode_getFullScreenModes
- sfRenderTexture_drawPrimitives
- sfRenderWindow_drawPrimitives 
- sfShader_setFloatUniformArray 
- sfShader_setVec2UniformArray 
- sfShader_setVec3UniformArray 
- sfShader_setVec4UniformArray 
- sfShader_setMat3UniformArray 
- sfShader_setMat4UniformArray 
- sfShape_create 
- sfText_findCharacterPos
- sfVertexArray_getVertexCount
- sfVertexArray_getVertex
- sfVertexArray_resize 
- sfSoundRecorder_create
- sfSoundRecorder_getAvailableDevices 

A few others use `ulong`
- sfFront_createFromMemory
- sfImage_createFromMemory
- sfTexture_createFromMemory
- sfMusic_createFromMemory
- sfSoundBuffer_createFromMemory

given that `size_t` size is platform dependent, I believe in C# you'd normally use `IntPtr`, `UIntPtr` or the newer `nint` or `nuint` types in place of `size_t`. I'm not sure in detail what the consequences are for the current mappings of `size_t` but I think this was worth raising going forward as .NET does have preferred types for platform dependent sized types.

PR #237, opened 2024-01-03
Unify all size_t arguments to use uint

DemoXinMC:
Somewhat addresses #212
* Updated to use `uint` consistently where CSFML is expecting `size_t`
* `UIntPtr` is the most correct type for `size_t`, but switching to it would result in breaking changes for end users
* uint should be replaced with `nuint` when moving to .NET 5+


eXpl0it3r:
Superseded by #238

PR #238, opened 2024-01-03
Unify all size_t arguments to use UIntPtr

DemoXinMC:
Addresses #212
* Updated to use a `SIZE_T` "typedef" where CSFML is expecting `size_t`
* `using SIZE_T = UIntPtr;` is at the beginning of each file that will P/Invoke functions with `size_t` signatures
* `SIZE_T` is currently typedef'd to `UIntPtr` but this can be changed when `nuint` becomes available to SFML.NET (.NET 5+)
* Since all changes are within protected/private code, this should not represent a breaking change for end-users

* Implemented Font(byte[]) properly using sfFont_createFromMemory() instead of creating a new MemoryStream

### DIFF
--- a/src/SFML.Audio/Music.cs
+++ b/src/SFML.Audio/Music.cs
@@ -4,6 +4,8 @@ using System.Runtime.InteropServices;
 using System.Security;
 using SFML.System;
 
+using SIZE_T = System.UIntPtr;
+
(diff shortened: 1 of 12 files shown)


eXpl0it3r:
Superseded by #253

Still appreciate the work! 🙂

Issue #195, opened 2020-10-31
VertexArray indexer doesnt returns a reference


jcbritobr:
This indexer in VertexArray does'nt returns a reference to a vertex. Docs are wrong, or the indexer is wrong. A deepy copy leads to poor performance in painting.

```csharp
//
        // Summary:
        //     Read-write access to vertices by their index. This function doesn't check index,
        //     it must be in range [0, VertexCount - 1]. The behaviour is undefined otherwise.
        //
        // Parameters:
        //   index:
        //     Index of the vertex to get
        //
        // Returns:
        //     Reference to the index-th vertex
        public Vertex this[uint index] { get; set; }
```

This code will not work because v is a Vertex copy.
```csharp
        Vertex v = array[(uint)i];
        v.Color = new Color((byte)r.Next(0, 255), (byte)r.Next(0, 255), (byte)r.Next(0, 255));
        v.Position = new Vector2f(r.Next(0, 800), r.Next(0, 600));
```

DangerRoss:
I'm gonna assume the docs are wrong on this detail as they look copied from the [C++ documentation](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1VertexArray.php#a6006676417d91f42d0278f1abcfe4352) where this indexer does return a reference. I'd also imagine the VertexArray class was originally written before return-by-reference for value types was introduced into C#.

jcbritobr:
> 
> 
> I'm gonna assume the docs are wrong on this detail as they look copied from the [C++ documentation](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1VertexArray.php#a6006676417d91f42d0278f1abcfe4352) where this indexer does return a reference. I'd also imagine the VertexArray class was originally written before return-by-reference for value types was introduced into C#.

Yep. My sugestion is to maintain indexer as it is, because ref can't work with this operator and create a
```csharp
public ref Vertex GetVertex(int index) {
    ...
}
```
This will boost the vertex painting performance a lot.

DangerRoss:
In this case, I don't believe the `Vertex` copy directly affects draw performance of a `VertexArray` as this indexer isn't used when drawing, but it obviously contributes to the time it takes to update individual vertexs. Any change for performance boosts is something I'd encorage benchmarking first.

Regarding your suggestion, I'd ask @DemoXinMC for input.

jcbritobr:
[...]

Issue #194, opened 2020-10-30
New Textures aren't zeroed


sjaak31367:
Hello!

I managed to stumble over a bug in a private project, and after quite a bit of digging and testing managed to conclude where it came from.
The bug was that generated images _sometimes_ contained semi-random pixels/images, even though that shouldn't be there.
After bug-hunting I discovered that
```cs
Texture texture = new Texture(uint sizeX, uint sizeY);
Image image = Texture.CopyToImage();
```
Did not produce an image where all pixels were zero (or equivalent).
  
\
On one hand this makes sense, as it gets assigned a chunk of memory, and if there's still data-remnants in that chunk, well, that's now part of its data.
On the other hand, I'd expect a new Texture to be empty; to contain nothing!

---
TL;DR: `Texture(uint sizeX, uint sizeY)` does not create an empty Texture.
This can be circumvented by doing:
```cs
Image image = new Image(sizeX, sizeY, new Color(0, 0, 0, 0));
Texture texture = new Texture(image);
```
But I'd argue that expecting a newly created Texture to be empty, is not unreasonable.

[Here](https://github.com/sjaak31367/SFML_nonzeroed_Texture_demonstration)'s a very small thingy to demonstrate that newly created Texures aren't empty. (Or possibly that `CopyToImage()` grabs data that should be zeroes.)

Greetings, and kind regards!
~Sjaak

LaurentGomila:
If the doc doesn't explicitly says that it should be zeroed, then it's not a bug. Moreover, as you said, this behavior makes sense.

So it's more a feature request actually 😉

But what's the point of defining the content of an uninitialized texture? Who's going to rely on that? If you want a black texture, then explicitly create one in your code.

eXpl0it3r:
The mentioned constructor simply calls the other mentioned constructor with the color black.
I'll have to try and reproduce this, maybe it's just some missing initialization somewhere.

https://github.com/SFML/SFML.Net/blob/43926e92e24750f364abbb2455b5033e3c226999/src/SFML.Graphics/Image.cs#L25-42

LaurentGomila:
@eXpl0it3r don't confuse between `Image` and `Texture`. What leaves the content uninitialized is to call `new Texture(width, height)`, without using an `Image`, and as I said it's fine like this.

eXpl0it3r:
Ah, I had the feeling I was missing something, but somehow forgot that original code was about the texture.
[...]

PR #239, opened 2024-01-03
In-code Documentation Update

DemoXinMC:
This is an absolute doozy of a diff, but it's all within comments and doesn't include any actual code changes.  It also adds clarity to a couple of quirks that are discussed in #194 and #195 

* Addresses requests in #194 and #195
* Updated all [Obsolete] attributes to be less repetitive
* Added a large number of quick reference links to existing summaries
* Corrected various typos and unparsable markup(\a and \p)
* Added additional formatting to numerous summaries, resulting in more readable tooltips
* Added a suggestion to use TimeSpan over SFML's Time while working within a managed environment

### DIFF
--- a/src/SFML.Graphics/RenderTexture.cs
+++ b/src/SFML.Graphics/RenderTexture.cs
@@ -34,7 +34,7 @@ namespace SFML.Graphics
         /// <param name="height">Height of the render-texture</param>
         /// <param name="depthBuffer">Do you want a depth-buffer attached?</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("Creating a RenderTexture with depthBuffer is deprecated. Use RenderTexture(width, height, contextSettings) instead.")]
+        [Obsolete("Use RenderTexture(width, height, contextSettings)")]
         public RenderTexture(uint width, uint height, bool depthBuffer) :
             base(sfRenderTexture_create(width, height, depthBuffer))
         {
@@ -525,8 +525,7 @@ namespace SFML.Graphics
         private Texture myTexture = null;
 
         #region Imports
-        [Obsolete("sfRenderTexture_create is obselete. Use sfRenderTexture_createWithSettings instead.")]
-        [DllImport(CSFML.graphics, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+        [DllImport(CSFML.graphics, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity, Obsolete]
         private static extern IntPtr sfRenderTexture_create(uint Width, uint Height, bool DepthBuffer);
 
         [DllImport(CSFML.graphics, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
--- a/src/SFML.Graphics/Shader.cs
+++ b/src/SFML.Graphics/Shader.cs
@@ -498,7 +498,7 @@ namespace SFML.Graphics
         /// <param name="x">Value to assign</param>
         ///
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, float x)
         {
             sfShader_setFloatParameter(CPointer, name, x);
@@ -516,7 +516,7 @@ namespace SFML.Graphics
         /// <param name="x">First component of the value to assign</param>
         /// <param name="y">Second component of the value to assign</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, float x, float y)
         {
             sfShader_setFloat2Parameter(CPointer, name, x, y);
@@ -535,7 +535,7 @@ namespace SFML.Graphics
         /// <param name="y">Second component of the value to assign</param>
         /// <param name="z">Third component of the value to assign</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, float x, float y, float z)
         {
             sfShader_setFloat3Parameter(CPointer, name, x, y, z);
@@ -555,7 +555,7 @@ namespace SFML.Graphics
         /// <param name="z">Third component of the value to assign</param>
         /// <param name="w">Fourth component of the value to assign</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, float x, float y, float z, float w)
         {
             sfShader_setFloat4Parameter(CPointer, name, x, y, z, w);
@@ -572,7 +572,7 @@ namespace SFML.Graphics
         /// <param name="name">Name of the parameter in the shader</param>
         /// <param name="vector">Vector to assign</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, Vector2f vector)
         {
             SetParameter(name, vector.X, vector.Y);
@@ -589,7 +589,7 @@ namespace SFML.Graphics
         /// <param name="name">Name of the parameter in the shader</param>
         /// <param name="color">Color to assign</param>
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
         public void SetParameter(string name, Color color)
         {
             sfShader_setColorParameter(CPointer, name, color);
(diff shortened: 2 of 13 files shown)


src/SFML.Graphics/RenderTexture.cs
```diff
@@ -525,8 +525,7 @@ protected override void Destroy(bool disposing)
         private Texture myTexture = null;
 
         #region Imports
-        [Obsolete("sfRenderTexture_create is obselete. Use sfRenderTexture_createWithSettings instead.")]
-        [DllImport(CSFML.graphics, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
+        [DllImport(CSFML.graphics, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity, Obsolete]
```
eXpl0it3r, line 528: Doesn't this remove the hint for `Use sfRenderTexture_createWithSettings instead.`?

src/SFML.Graphics/Shader.cs
```diff
@@ -498,7 +498,7 @@ public unsafe void SetUniformArray(string name, Glsl.Mat4[] array)
         /// <param name="x">Value to assign</param>
         ///
         ////////////////////////////////////////////////////////////
-        [Obsolete("SetParameter is deprecated, please use the corresponding SetUniform")]
+        [Obsolete("Use SetUniform()")]
```
eXpl0it3r, line 501: Wonder if we should keep the `the corresponding` since there are matching overloads

src/SFML.Graphics/Text.cs
```diff
@@ -97,18 +97,20 @@ public enum Styles
 
         ////////////////////////////////////////////////////////////
         /// <summary>
-        /// Fill color of the object
+        /// DEPRECATED <see cref="SFML.Graphics.Color"/> of the <see cref="Text"/>
```
eXpl0it3r, line 100: The `Fill` was dropped

src/SFML.Graphics/Text.cs
```diff
@@ -281,14 +287,14 @@ public Vector2f FindCharacterPos(uint index)
...
-        /// The returned rectangle is in local coordinates, which means
-        /// that it ignores the transformations (translation, rotation,
-        /// scale, ...) that are applied to the entity.
-        /// In other words, this function returns the bounds of the
-        /// entity in the entity's coordinate system.
+        /// Get the LOCAL bounding <see cref="FloatRect"/> of the text.
         /// </summary>
+        /// <remarks>
+        /// The returned <see cref="FloatRect"/> is in LOCAL coordinates.
+        /// <para>Transformations (Translation, Rotation, Scale) ARE NOT applied to the entity.</para>
```
eXpl0it3r, line 294: I think `are not` (lower case) is sufficient here

src/SFML.Graphics/Text.cs
```diff
@@ -298,14 +304,14 @@ public FloatRect GetLocalBounds()
...
-        /// The returned rectangle is in global coordinates, which means
-        /// that it takes in account the transformations (translation,
-        /// rotation, scale, ...) that are applied to the entity.
-        /// In other words, this function returns the bounds of the
-        /// sprite in the global 2D world's coordinate system.
+        /// Get the GLOBAL bounding rectangle of the text.
         /// </summary>
+        /// <remarks>
+        /// The returned <see cref="FloatRect"/> is in GLOBAL coordinates.
+        /// <para>Transformations (Translation, Rotation, Scale) ARE applied to the entity.</para>
```
eXpl0it3r, line 311: Same here: `are`

src/SFML.Graphics/VertexArray.cs
```diff
@@ -111,13 +112,14 @@ public void Clear()
         ////////////////////////////////////////////////////////////
         /// <summary>
         /// Resize the vertex array
-        /// 
-        /// If \a vertexCount is greater than the current size, the previous
+        /// </summary>
+        /// <remarks>
+        /// If vertexCount is greater than the current size, the previous
```
eXpl0it3r, line 117:
```suggestion
        /// If <see cref="VertexCount"/> is greater than the current size, the previous
```

src/SFML.Graphics/VertexArray.cs
```diff
@@ -111,13 +112,14 @@ public void Clear()
...
         /// vertices are kept and new (default-constructed) vertices are
         /// added.
-        /// If \a vertexCount is less than the current size, existing vertices
+        /// If vertexCount is less than the current size, existing vertices
```
eXpl0it3r, line 120:
```suggestion
        /// If <see cref="VertexCount"/> is less than the current size, existing vertices
```

src/SFML.Graphics/VertexBuffer.cs
```diff
@@ -177,27 +187,26 @@ public bool Update(Vertex[] vertices, uint vertexCount, uint offset)
...
+        /// <para>If the size of the <see cref="Vertex"/>[] is greater than the
+        /// current size of the buffer, a new buffer is created containing
+        /// the <see cref="Vertex"/> data.</para>
         ///
-        /// If offset is not 0 and offset + vertexCount is greater
-        /// than the size of the currently created buffer, the update fails.
+        /// <para>If the size of the <see cref="Vertex"/>[] is less than the
+        /// current size of the buffer, only the first Vertex[].Length vertices
+        /// are replaced.</para>
+        /// </remarks>
```
eXpl0it3r, line 203: `</remarks>` closed too early

eXpl0it3r (changes requested):
Love the PR! 🙂 
A few minor adjustments

src/SFML.Graphics/RenderTexture.cs
DemoXinMC: It does remove that hint, but only for calling `sfRenderTexture_create()` directly.  Since that DllImport is private, then end user would never see that warning anyways. The warning still exists on the public-facing API.

src/SFML.Graphics/Shader.cs
DemoXinMC: There are matching overloads that don't specify the type of argument they take, so the end-user would be calling `.SetUniform(whateverTypeOfUniform)` and shouldn't be prompted to look for another method that corresponds to a specific argument type.

src/SFML.Graphics/Text.cs
DemoXinMC: LOCAL and ARE NOT were capitalized to emphasize the differences between GetLocalBounds() and GetGlobalBounds().  Having emphasis for details like this isn't out of the ordinary from my experience, but if we want to remove that emphasis, I completely understand and can make that change.

src/SFML.Graphics/Text.cs
DemoXinMC: Same as GetLocalBounds():
GLOBAL and ARE were capitalized to emphasize the differences between GetLocalBounds() and GetGlobalBounds().  Having emphasis for details like this isn't out of the ordinary from my experience, but if we want to remove that emphasis, I completely understand and can make that change.

src/SFML.Graphics/VertexArray.cs
DemoXinMC: Great catch

DemoXinMC: I don't know that this works/has a purpose.  The <param> tag provides information about the vertexCount argument.

src/SFML.Graphics/VertexArray.cs
DemoXinMC: I don't know that this works/has a purpose.  The <param> tag provides information about the vertexCount argument.

src/SFML.Graphics/VertexBuffer.cs
DemoXinMC: Good catch

src/SFML.Graphics/Text.cs
DemoXinMC: I'll add a bit more clarity to this.  People should be moving away from `Text.Color` though, rather than getting better documentation for it. 🤣

src/SFML.Graphics/VertexArray.cs
DemoXinMC: Upon further investigation, it appears that <paramref name="vertexCount" /> would be the correct method of doing this.  This tag could be used in a LOT of places.

src/SFML.Graphics/Text.cs
eXpl0it3r: Yeah not a fan and I think it's not necessary.

DemoXinMC: Roger.  I'll go ahead and nix these from both then,

DemoXinMC:
I've updated the PR with some more changes.

* Adjusted changes to align with eXpl0it3r's feedback.
* Added a compile-time #warning to `VertexBuffer.NativeHandle`'s getter warning the user to be cautious with `NativeHandle` and that most people won't need to use it.
* Made use of `<paramref />` in many places.  Lots more of this to do.

src/SFML.Graphics/Shader.cs
eXpl0it3r: I guess we can keep it as is

src/SFML.Graphics/VertexBuffer.cs
```diff
@@ -88,20 +91,24 @@ public VertexBuffer(VertexBuffer copy)
...
         /// very specific stuff to implement that SFML doesn't support,
         /// or implement a temporary workaround until a bug is fixed.
-        /// </summary>
+        /// </remarks>
         ////////////////////////////////////////////////////////////
         public uint NativeHandle
         {
-            get { return sfVertexBuffer_getNativeHandle(CPointer); }
+            get {
+                #warning Use NativeHandle cautiously.  Most people shouldn't need it.
```
eXpl0it3r, line 104: Not sure about the warning here. Can one turn it off? It seems to get triggered on the build itself, see the annotations below from every build.

If you really want to keep it, remove the double space between the two sentences.

eXpl0it3r (approved):

eXpl0it3r:
Thank you for these updates! 🙂

PR #243, opened 2024-02-16
Implement most CSFML 2.6.0 features

Marioalexsan:
This PR aims to implement most things from CSFML 2.6.0, along with some stuff that might be missing from CSFML 2.5.2.

Change list from 2.6.0:
- `sfFont_setSmooth` / `sfFont_isSmooth`
- `sfBlendMin` & `sfBlendMax`
- `sfFloatRect_getPosition`/`sfIntRect_getPosition` & `sfFloatRect_getSize`/`sfIntRect_getSize`
- Document MP3 support
- Add directional arrow resize cursors
- `sfFont_getBoldKerning`
- Add `sf::Vulkan` functions (partial)
- Add scancode support (partial)
- `sfRenderTexture_isSrgb` / `sfRenderWindow_isSrgb`
- `sfImage_saveToMemory` with `sfBuffer`
- `sf::WindowBase`

Other things which were done in this PR (probably missing from CSFML 2.5.2?):
- `drawVertexBufferRange` implementations from CSFML
- `TextureCoordinateType` implementation
- sRGB creation functions
- `sfContext_isExtensionAvailable` and `sfContext_getFunction`
- `sfWindowBase_createVulkanSurface`
- some other stuff I probably missed

Features which were not done due to issues:
- `sfVulkan_getGraphicsRequiredInstanceExtensions` - needs a CSFML release containing [CSFML#229](https://github.com/SFML/CSFML/pull/229)
- `sfKeyboard_getDescription` - this function returns an owning C pointer (`const char*`). I'm not sure how to handle the deallocation for it since there's no corresponding function to free the string

By the looks of it, this PR might overlap / conflict with #236, #239, #240, and maybe #238.

This PR is a mess and I don't expect it to pass first inspection, so any feedback is welcome.

### DIFF
--- a/src/SFML.Window/WindowBase.cs
+++ b/src/SFML.Window/WindowBase.cs
@@ -1,0 +1,797 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Text;
+using SFML.System;
+
+namespace SFML.Window
+{
+    ////////////////////////////////////////////////////////////
+    /// <summary>
+    /// Enumeration of window creation styles
+    /// </summary>
+    ////////////////////////////////////////////////////////////
+    [Flags]
+    public enum Styles
+    {
+        /// <summary>No border / title bar (this flag and all others are mutually exclusive)</summary>
+        None = 0,
+
+        /// <summary>Title bar + fixed border</summary>
+        Titlebar = 1 << 0,
+
+        /// <summary>Titlebar + resizable border + maximize button</summary>
+        Resize = 1 << 1,
+
+        /// <summary>Titlebar + close button</summary>
+        Close = 1 << 2,
+
+        /// <summary>Fullscreen mode (this flag and all others are mutually exclusive))</summary>
+        Fullscreen = 1 << 3,
+
+        /// <summary>Default window style (titlebar + resize + close)</summary>
+        Default = Titlebar | Resize | Close
+    }
+
+
+    ////////////////////////////////////////////////////////////
+    /// <summary>
+    /// Window that serves as a base for other windows
+    /// </summary>
+    ////////////////////////////////////////////////////////////
+    public class WindowBase : ObjectBase
+    {
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Create the window with default style and creation settings
+        /// </summary>
+        /// <param name="mode">Video mode to use</param>
+        /// <param name="title">Title of the window</param>
+        ////////////////////////////////////////////////////////////
+        public WindowBase(VideoMode mode, string title) :
+            this(mode, title, Styles.Default)
+        {
+        }
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Create the window with default creation settings
+        /// </summary>
+        /// <param name="mode">Video mode to use</param>
+        /// <param name="title">Title of the window</param>
+        /// <param name="style">Window style (Resize | Close by default)</param>
+        ////////////////////////////////////////////////////////////
+        public WindowBase(VideoMode mode, string title, Styles style) :
+            base(IntPtr.Zero)
+        {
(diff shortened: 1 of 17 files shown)


Marioalexsan:
Rebased from master

src/SFML.Window/WindowBase.cs
```diff
@@ -0,0 +1,797 @@
...
+        /// </summary>
+        /// <param name="mode">Video mode to use</param>
+        /// <param name="title">Title of the window</param>
+        /// <param name="style">Window style (Resize | Close by default)</param>
+        ////////////////////////////////////////////////////////////
+        public WindowBase(VideoMode mode, string title, Styles style) :
+            base(IntPtr.Zero)
+        {
+            // Copy the title to a null-terminated UTF-32 byte array
+            byte[] titleAsUtf32 = Encoding.UTF32.GetBytes(title + '\0');
```
eXpl0it3r, line 70:
```suggestion
            var titleAsUtf32 = Encoding.UTF32.GetBytes(title + '\0');
```

src/SFML.Window/WindowBase.cs
```diff
@@ -0,0 +1,797 @@
...
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Change the title of the window
+        /// </summary>
+        /// <param name="title">New title</param>
+        ////////////////////////////////////////////////////////////
+        public virtual void SetTitle(string title)
+        {
+            // Copy the title to a null-terminated UTF-32 byte array
+            byte[] titleAsUtf32 = Encoding.UTF32.GetBytes(title + '\0');
```
eXpl0it3r, line 148:
```suggestion
            var titleAsUtf32 = Encoding.UTF32.GetBytes(title + '\0');
```

src/SFML.Window/WindowBase.cs
```diff
@@ -0,0 +1,797 @@
...
+        /// <summary>
+        /// Constructor for derived classes
+        /// </summary>
+        /// <param name="cPointer">Pointer to the internal object in the C API</param>
+        /// <param name="dummy">Internal hack :)</param>
+        ////////////////////////////////////////////////////////////
+        protected WindowBase(IntPtr cPointer, int dummy) :
+            base(cPointer)
+        {
+            // TODO : find a cleaner way of separating this constructor from WindowBase(IntPtr handle)
```
eXpl0it3r, line 360: Can you describe what the issue currently is?

eXpl0it3r (changes requested):
Some small `var` changes

src/SFML.Window/WindowBase.cs
Marioalexsan: This is pretty much the same issue replicated from SFML.Window to SFML.WindowBase. The constructor for initializing a window from an OS handle has the same signature as the one to create a Window object from a CSFML pointer (i.e. both use IntPtr), so the extra dummy parameter is used to differentiate them. #240 might be an option to fix it, though.

DemoXinMC:
Is there some benefit to wrapping WindowBase that I'm missing? It's not extensible from C# as it's basically used as an interface in SFML.

Also, I realize that `var` is pretty acceptable to use, but we're using explicit `IntPtr` throughout most of the wrapping IIRC. Does it make sense to switch to `var` right now on this PR?

eXpl0it3r:
> Is there some benefit to wrapping WindowBase that I'm missing? It's not extensible from C# as it's basically used as an interface in SFML.

What do you mean with "wrapping". The point of `WindowBase` is to create a window without OpenGL context.  
It's probably not as useful since it's tricky to use pure OpenGL or Vulkan in C#

> Also, I realize that `var` is pretty acceptable to use, but we're using explicit `IntPtr` throughout most of the wrapping IIRC. Does it make sense to switch to `var` right now on this PR?

I wouldn't change it on this PR, it's already mixing a lot of things.

eXpl0it3r (approved):
LGTM

eXpl0it3r:
Thank you for implementing this all! 🙂

PR #244, opened 2024-02-25
Binding and API fixes

Marioalexsan:
Fixed binding signatures (which were originally broken by yours truly):
```
CSFML_WINDOW_API sfBool sfContext_isExtensionAvailable(const char* name);
CSFML_WINDOW_API GlFunctionPointer sfContext_getFunction(const char* name);
```

Removed old bindings that don't exist anymore in CSFML:
```
private static extern bool sfRenderTexture_saveGLStates(IntPtr CPointer);
private static extern bool sfRenderTexture_restoreGLStates(IntPtr CPointer);
private static extern bool sfRenderWindow_saveGLStates(IntPtr CPointer);
private static extern bool sfRenderWindow_restoreGLStates(IntPtr CPointer);
private static extern uint sfRenderWindow_getFrameTime(IntPtr CPointer);
private static extern uint sfWindow_getFrameTime(IntPtr CPointer);
private static extern IntPtr sfShape_copy(IntPtr Shape);
private static extern FloatRect sfText_getRect(IntPtr CPointer);
private static extern FloatRect sfTexture_getTexCoords(IntPtr texture, IntRect rectangle);
```

Fixed missing userdata pointers in SoundRecorder callbacks.

Moved `sfMouse_getPosition`, `sfMouse_setPosition`, and `sfTouch_getPosition` back to `Window`, since those functions operate on `sfWindow` structs and not `sfWindowBase`. `WindowBase` will need its own CSFML functions that are not available at the moment, which is why I used `NotImplementedException` for the time being.

Also changed `Mouse` and `Touch` functions to use `WindowBase` instead of `Window`.

### DIFF
--- a/src/SFML.Window/WindowBase.cs
+++ b/src/SFML.Window/WindowBase.cs
@@ -394,7 +394,7 @@ namespace SFML.Window
         ////////////////////////////////////////////////////////////
         protected internal virtual Vector2i InternalGetMousePosition()
         {
-            return sfMouse_getPosition(CPointer);
+            throw new NotImplementedException("Currently not available");
         }
 
         ////////////////////////////////////////////////////////////
@@ -407,7 +407,7 @@ namespace SFML.Window
         ////////////////////////////////////////////////////////////
         protected internal virtual void InternalSetMousePosition(Vector2i position)
         {
-            sfMouse_setPosition(position, CPointer);
+            throw new NotImplementedException("Currently not available");
         }
 
         ////////////////////////////////////////////////////////////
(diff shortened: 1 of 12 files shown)


src/SFML.Window/WindowBase.cs
```diff
@@ -421,7 +421,7 @@ protected internal virtual void InternalSetMousePosition(Vector2i position)
         ////////////////////////////////////////////////////////////
         protected internal virtual Vector2i InternalGetTouchPosition(uint Finger)
         {
-            return sfTouch_getPosition(Finger, CPointer);
+            throw new NotImplementedException("Currently not available");
```
eXpl0it3r, line 424: Are we not implementing it, because there hasn't been a new CSFML release yet with that function included?

eXpl0it3r: These internal functions exist because the getPosition overloads for a window, needs the private window CPointer.

Additionally, this is not implemented here, because we haven't created a CSFML 2.6.1 release with the missing function implementation.

eXpl0it3r (approved):

eXpl0it3r:
Took a while, but finally got around to it.

For SFML.Net 2.6.0, I guess we still have to release CSFML 2.6.1, so we can implement the WindowBase functions.

Issue #245, opened 2024-04-20
Dispatch of `KeyPress` event throws a `System.AccessViolationException`


AzuxirenLeadGuy:
# Issue description

While running a `RenderWindow` instance, any `KeyPress` event results in throwing a `System.AccessViolationException`.

# How to reproduce

Run the example project in `example/window`, and press any key

## Expected behaviour

The key is processed by the event successfully

## Actual behaviour

An uncaught exception interrupts the program, with the following message on the console.
```
Press ESC key to close window
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SFML.Window.WindowBase.CallEventHandler(SFML.Window.Event)
   at SFML.Window.WindowBase.DispatchEvents()
   at window_core.SimpleWindow.Run()
   at window_core.Program.Main(System.String[])
```

eXpl0it3r:
It might be recommended to stick to CSFML 2.5.2 until SFML.Net has been updated for 2.6

eXpl0it3r:
Note, that while the change has been merged into master, it doesn't mean that you can now mix the SFML.Net 2.5.x nuget packages with the CSFML 2.6.x nuget packages.

Keep the versions aligned to not run into this issue. 😉

AzuxirenLeadGuy:
> Note, that while the change has been merged into master, it doesn't mean that you can now mix the SFML.Net 2.5.x nuget packages with the CSFML 2.6.x nuget packages.

Just to clarify, this won't happen unless there are multiple SFML.NET projects being worked on, right? If I only use the latest SFML.Net nupkg (built from the current master branch), I don't need to worry about this, right?

eXpl0it3r:
If you build SFML.Net yourself, then you should be okay for this issue.

It's just that there's no official SFML.Net 2.6 Nuget package yet, and people seem to be trigger happy on the update button for CSFML 😄

uwaaya:
in our case, this actually had nothing to do with nuget. upstream changed the CSFML version, and we just ran into this issue bc we just assumed upstream wouldnt break their master branch. its no big deal though. ppl simply gotta reset to a stable tag after creating their fork, but most would not be aware of that..

this is just my thoughts, but i think it be a good idea to have a development branch for updating CSFML, right? i know its not needed most of the time, but SFML will change a lot with SFML 3 being in active development, and keeping master stable would be super helpful for anybody creating and maintaining their own forks :eyes:

[...]

Issue #226, opened 2023-05-02
Consider using Native Library Loading


eXpl0it3r:
Using p/invoke with `DllImportAttribute` is a bit cumbersome in a [cross-platform manner](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform).

As such, it might make sense to use the [NativeLibrary class](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary?view=net-6.0) instead.

## References

- https://learn.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform
- https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary?view=net-6.0
- https://github.com/dotnet/runtime/issues/27267
- https://github.com/dotnet/runtime/issues/4216

DemoXinMC:
I'm a bit confused on the benefit of this.

In addition to the large scale rewrite of P/Invoke definitions, wouldn't this also either require users to manually call some static method we create like `SFML.Modules.LoadLibrary()`(name just made up) or add their own `NativeLibrary.Load("sfml-system.dll");` at the start of their program?

eXpl0it3r:
Having to workaround limitations of `DllImport` again (see #247), I seem to understand the underlying problem a bit better.

We can provide a single name to `DllImport` and then have to hope that InteropServices does the correct thing in resolving the libraries cross-platform. As an example for `csfml-graphics` it would need to find on Windows `csfml-graphics.dll`, on Linux `libcsfml-graphics.so`, and on macOS `libcsfml-graphics.dylib`, but that in a specific directory, that is the `bin/<Config>/<TargetFramework>/runtimes/<RID>/native/`.

While it actually gets pretty close, looks in all those directories, it fails to look for the right combination - in my opinion a bug. Thus it forces us to ship libraries without suffix, e.g. `libcsfml-graphics` on Linux and macOS.

There's one potential workaround and that is to use write our own Custom Import Resolver, so we can define ourselves what is being looked for: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading#custom-import-resolver

The use of Native Library Loading was just another idea, that I don't fully understand (yet) and might not be suitable.

eXpl0it3r:
[...]

PR #247, opened 2024-05-18
Use major and minor version in the CSFML library names

eXpl0it3r:
This solves two issues:

- For one it ensures that we're not using old or new versions of CSFML (e.g. #245) that may not be fully compatible
- It will also allow us to drop duplicated library versions in the CSFML nuget (no sfml-graphics.so or sfml-graphics.2.6.1.so).

### DIFF
--- a/src/SFML.Audio/Listener.cs
+++ b/src/SFML.Audio/Listener.cs
@@ -69,28 +69,28 @@ namespace SFML.Audio
         }
 
         #region Imports
-        [DllImport(CSFML.audio, CallingConvention = CallingConvention.Cdecl), SuppressUnmanagedCodeSecurity]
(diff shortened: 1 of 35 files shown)


eXpl0it3r:
A limitation is, that it won't find default CSFML library names anymore, meaning on Linux you'd get `libsfml-graphics.so.2.6`, but now SFML.Net is looking for `libsfml-graphics.2.6.so`, nor for Windows where it's `csfml-graphics.dll`, it now needs to be `csfml-graphics.2.6.dll`.

I wonder if it's worth doing this hardcoding, or if we should just ship the ones needed. That would `libsfml-*.so`, `libsfml-*.so.2.6`, `libsfml-*.dylib`, and `libsfml-*.2.6.dylib`

eXpl0it3r:
After the conclusions in #226, we won't go with this change.

PR #250, opened 2024-05-22
Add missing Vulkan and Keyboard functions

Marioalexsan:
See https://github.com/SFML/CSFML/pull/236

### DIFF
--- a/src/SFML.Graphics/RenderWindow.cs
+++ b/src/SFML.Graphics/RenderWindow.cs
@@ -331,6 +331,20 @@ namespace SFML.Graphics
 
         ////////////////////////////////////////////////////////////
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/SFML.System/SfAlloc.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Security;
+using System;
+
+namespace SFML.System
+{
+    /// <summary>
+    /// Contains functions related to memory allocation.
+    /// For internal use only.
+    /// </summary>
+    public static class SfAlloc
```
eXpl0it3r, line 11: Why not just `Allocation`?

eXpl0it3r (approved):

eXpl0it3r:
Thank you for the follow up! 🙂

PR #252, opened 2024-05-27
Update sRGB API for Texture and RenderTarget

### LINKED ISSUE #222: Add proper support for sRGB

Marioalexsan:
Fixes #222, supersedes #236.

As far as I'm aware, most of the functionality for sRGB was already added through other PRs.


src/SFML.Graphics/RenderTarget.cs
```diff
@@ -1,3 +1,4 @@
+using System.Reflection;
```
eXpl0it3r, line 1: Don't think we need Reflection here.

src/SFML.Graphics/RenderTarget.cs
```diff
@@ -16,6 +17,14 @@ public interface RenderTarget
         ////////////////////////////////////////////////////////////
         Vector2u Size { get; }
 
+
+        ////////////////////////////////////////////////////////////
+        /// <summary>
+        /// Tell if the render target will use sRGB encoding when drawing on it
+        /// </summary>
+        ////////////////////////////////////////////////////////////
+        bool IsSrgb { get; }
```
eXpl0it3r, line 26: Where is this set?

eXpl0it3r (changes requested):

src/SFML.Graphics/RenderTarget.cs
Marioalexsan: Removed it.

src/SFML.Graphics/RenderTarget.cs
eXpl0it3r: Ah my bad, I didn't realized this is on RenderTarget and that RenderTarget is an interface 😄

eXpl0it3r (approved):

eXpl0it3r:
Thank you for aligning this! 🙂

PR #253, opened 2024-05-27
Use UIntPtr for size_t arguments and return values from CSFML

Marioalexsan:
Supersedes #238.

Most of the code is the same as DemoXin's, but I used `UIntPtr` directly and updated it with the remaining places that use `size_t` in CSFML.

### DIFF
--- a/src/SFML.Graphics/Font.cs
+++ b/src/SFML.Graphics/Font.cs
@@ -41,8 +41,10 @@ namespace SFML.Graphics
         ////////////////////////////////////////////////////////////
         public Font(Stream stream) : base(IntPtr.Zero)
         {
-            myStream = new StreamAdaptor(stream);
-            CPointer = sfFont_createFromStream(myStream.InputStreamPtr);
+            using (var adaptor = new StreamAdaptor(stream))
+            {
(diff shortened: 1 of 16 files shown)


src/SFML.Graphics/Font.cs
```diff
@@ -57,7 +59,23 @@ public Font(Stream stream) : base(IntPtr.Zero)
...
+                CPointer = sfFont_createFromMemory(pin.AddrOfPinnedObject(), (UIntPtr)bytes.Length);
+            }
+            finally
+            {
+                pin.Free();
+            }
+            if (CPointer == IntPtr.Zero)
+            {
+                throw new LoadingFailedException("font");
+            }
```
eXpl0it3r, line 77: This is change from createFromStream to createFromMemory do anything for us?

Marioalexsan: This way the data is read directly from memory instead of having to make a `MemoryStream`, `StreamAdaptor` and associated callback stuff. It's also more consistent with the other places in the code that have `createFromStream` and `createFromMemory`.

eXpl0it3r (approved):

eXpl0it3r:
Thank you! 🙂

PR #254, opened 2024-05-30
Fix AccessViolationException on KeyPress

### LINKED ISSUE #245: Dispatch of `KeyPress` event throws a `System.AccessViolationException`

nulldg:
The scancode field had been added to the key event in CSFML and this change was brought downstream in c68ed27. Master has been broken ever since because there was never any followup to update the binding.
It's a very simple bug and it's a severe one, and the exact cause has been known about for weeks, so why the fuck is this taking so long to address? Anyway, hope this helps.

Closes #245

### DIFF
--- a/src/SFML.Window/Event.cs
+++ b/src/SFML.Window/Event.cs
@@ -92,6 +92,9 @@ namespace SFML.Window
         /// <summary>Code of the key. See <see cref="Keyboard.Key"/></summary>
         public Keyboard.Key Code;
 
+        /// <summary>Physical code of the key. See <see cref="Keyboard.Scancode"/></summary>
+        public Keyboard.Scancode Scancode;
+
         /// <summary>Is the Alt modifier pressed?</summary>
         public int Alt;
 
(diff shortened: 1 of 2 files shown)


Marioalexsan (approved):

ChrisThrasher:
Asking us why the fuck we’re so slow isn’t helping merge your PR faster. Please reconsider your language.

eXpl0it3r:
> It's a genuine question [...] this crash bug isn't taken seriously

If you want to be taken seriously, you may want to use non-aggressive language and manage your emotional balance.

SFML.Net 2.6.0 nor 2.6.1 have not been released and as such there's no guarantee on the completeness.  
Additionally, if you use CSFML 2.6.0 or CSFML 2.6.1 with SFML.Net 2.5.0 or 2.5.1, you're missing some understanding of how SFML.Net depends on CSFML which depends on SFML and that the versions have to align.

Don't just press that update button, without understanding the impact.

And last but not least, nobody here owes anyone, anything. We're all working on this in our spare time and out of our own interest, not because someone has the "right" to a constantly up to date version.

eXpl0it3r (approved):

PR #258, opened 2024-06-06
Use fixed version for CSFML

Marioalexsan:
This theoretically prevents users from accidentally updating CSFML to a newer version from the NuGet Updates tab. It doesn't stop them from forcefully installing a newer version of CSFML as per <https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution#direct-dependency-wins>.

### DIFF
--- a/SFML.Module.props
+++ b/SFML.Module.props
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="CSFML" Version="2.6.1" />
+    <PackageReference Include="CSFML" Version="[2.6.1, 2.7)" />
   </ItemGroup>
 
 </Project>


SFML.Module.props
```diff
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="CSFML" Version="2.6.1" />
+    <PackageReference Include="CSFML" Version="[2.6.1]" />
```
eXpl0it3r, line 20: Should this be limited up to the patch version?

In SFML.Net 2.5.* it was useful to update CSFML until we provided a new SFML.Net version

eXpl0it3r (approved):

Issue #251, opened 2024-05-23
SFML.Net 2.6.0 Release


eXpl0it3r:
Listing a the things that still need to happen before we can release SFML.Net 2.6.0

- [x] #249
  - [x] Use CSFML 2.6.1
  - [x] WindowBase.InternalGetMousePosition 
  - [x] WindowBase.InternalSetMousePosition
  - [x] WindowBase.InternalGetTouchPosition
- [x] #250
  - [x] Vulkan.GetGraphicsRequiredInstancExtensions
  - [x] KeyboardGetDescription (#250)
- [x] Finalize #222
  - [x] #236
  - [x] #252
- [x] Decide on
  - [x] #238
  - [x] #253
- [x] #191
- [x] Constrain #255
- [x] Add platform specifier to NetFX examples
- [x] Fix documentation
  - [x] renderimage isn't a thing: https://github.com/SFML/SFML.Net/blob/master/src/SFML.Graphics/RenderTarget.cs#L7C9-L7C75
  - [x] Mention that RenderTarget cannot be extended on custom types (see #259)
- [x] Update version in props & examples
- [x] Finalize [Release Check List](https://github.com/SFML/SFML/wiki/SD:-Releases#sfmlnet-release-check-list)

eXpl0it3r:
SFML.Net 2.6.0 has been released 

https://github.com/SFML/SFML.Net/releases/tag/2.6.0

PR #261, opened 2024-06-07
Fix documentation, up versions, change platform configuration for examples

Marioalexsan:
Pretty much leftover items from #251.

Not 100% sure if the platform configuration changes are it, but it prevents people from building for AnyCPU on .NET Framework projects. They can still try running the projects, and VS will fail to do so, but I'm not sure if there's something that can be done about that.

Also changes the platform configurations so that x64 and x86 actually build for x64 and x86 and not for AnyCPU.


eXpl0it3r (approved):

eXpl0it3r:
Thank you @Marioalexsan for getting this all cleaned up and getting SFML.Net 2.6.0 release ready! 🙂

PR #263, opened 2024-06-18
Use Span for SFML.Net API

Marioalexsan:
Making this draft as a starting point for Span-related changes. This might also be a good place to look over some memory allocations happening in the library. The idea is as follows:
- Use System.Memory to gain access to Span<T>, which is usable in .NET Standard 2.0
- Replace T[] with Span<T> in input parameters where possible - this is a trivial change that just allows slicing
- Replace T[] returns with Span<T> where possible (i.e. when we want to just give a view into memory) - this reduces the number of allocations and solves some anti-patterns like `Image.Pixels`

In some cases it's harder to use Span<T> due to lack of methods that are available only in .NET Core. One such example is StreamAdaptor where you cannot read into a Span<byte>.

I'm also thinking we could use `stackalloc` for places that use patterns like `byte[] titleAsUtf32 = Encoding.UTF32.GetBytes(title + '\0')`. There is an overload for `GetBytes` that can work with pointers, which could be used together with `stackalloc` for small-sized strings to avoid allocating small temporary strings on the heap.

### DIFF
--- a/src/SFML.Audio/SoundBuffer.cs
+++ b/src/SFML.Audio/SoundBuffer.cs
@@ -69,18 +69,17 @@ namespace SFML.Audio
         /// <param name="bytes">Byte array containing the file contents</param>
         /// <exception cref="LoadingFailedException" />
         ////////////////////////////////////////////////////////////
-        public SoundBuffer(byte[] bytes) :
+        public SoundBuffer(Span<byte> bytes) :
             base(IntPtr.Zero)
         {
-            GCHandle pin = GCHandle.Alloc(bytes, GCHandleType.Pinned);
-            try
+            unsafe
             {
-                CPointer = sfSoundBuffer_createFromMemory(pin.AddrOfPinnedObject(), (UIntPtr)bytes.Length);
+                fixed (byte* ptr = bytes)
+                {
+                    CPointer = sfSoundBuffer_createFromMemory((IntPtr)ptr, (UIntPtr)bytes.Length);
+                }
             }
-            finally
-            {
-                pin.Free();
-            }
+
             if (CPointer == IntPtr.Zero)
             {
                 throw new LoadingFailedException("sound buffer");
@@ -96,7 +95,7 @@ namespace SFML.Audio
         /// <param name="sampleRate">Sample rate</param>
         /// <exception cref="LoadingFailedException" />
         ////////////////////////////////////////////////////////////
-        public SoundBuffer(short[] samples, uint channelCount, uint sampleRate) :
+        public SoundBuffer(Span<short> samples, uint channelCount, uint sampleRate) :
             base(IntPtr.Zero)
         {
             unsafe
@@ -181,13 +180,17 @@ namespace SFML.Audio
         /// (sf::Int16).
         /// </summary>
         ////////////////////////////////////////////////////////////
-        public short[] Samples
+        public Span<short> Samples
         {
             get
             {
-                short[] SamplesArray = new short[sfSoundBuffer_getSampleCount(CPointer)];
-                Marshal.Copy(sfSoundBuffer_getSamples(CPointer), SamplesArray, 0, SamplesArray.Length);
-                return SamplesArray;
+                ulong count = sfSoundBuffer_getSampleCount(CPointer);
+                IntPtr ptr = sfSoundBuffer_getSamples(CPointer);
+
+                unsafe
+                {
+                    return new Span<short>((void*)ptr, (int)count);
+                }
             }
         }
 
--- a/src/SFML.Audio/SoundBufferRecorder.cs
+++ b/src/SFML.Audio/SoundBufferRecorder.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 namespace SFML.Audio
@@ -57,9 +58,14 @@ namespace SFML.Audio
         /// <param name="samples">Array of samples to process</param>
         /// <returns>False to stop recording audio data, true to continue</returns>
         ////////////////////////////////////////////////////////////
-        protected override bool OnProcessSamples(short[] samples)
+        protected override bool OnProcessSamples(Span<short> samples)
         {
-            mySamplesArray.AddRange(samples);
+            if (mySamplesArray.Capacity < mySamplesArray.Count + samples.Length)
+                mySamplesArray.Capacity = mySamplesArray.Count + samples.Length;
+
+            for (int i = 0; i < samples.Length; i++)
+                mySamplesArray.Add(samples[i]);
+
             return true;
         }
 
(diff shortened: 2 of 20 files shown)


src/SFML.Graphics/RenderTexture.cs
```diff
@@ -373,44 +373,15 @@ public void Draw(Vertex[] vertices, PrimitiveType type)
...
-        public void Draw(Vertex[] vertices, uint start, uint count, PrimitiveType type, RenderStates states)
+        public void Draw(Span<Vertex> vertices, PrimitiveType type, RenderStates states)
         {
             RenderStates.MarshalData marshaledStates = states.Marshal();
 
             unsafe
             {
                 fixed (Vertex* vertexPtr = vertices)
                 {
-                    sfRenderTexture_drawPrimitives(CPointer, vertexPtr + start, (UIntPtr)count, type, ref marshaledStates);
```
DemoXinMC, line 413: Do we want to *FORCE* using slices?  This removes the ability to specify offsets and lengths during `Draw()`

src/SFML.Graphics/RenderWindow.cs
```diff
@@ -184,13 +184,13 @@ public override void SetTitle(string title)
         /// <param name="height">Icon's height, in pixels</param>
         /// <param name="pixels">Array of pixels, format must be RGBA 32 bits</param>
         ////////////////////////////////////////////////////////////
-        public override void SetIcon(uint width, uint height, byte[] pixels)
+        public override void SetIcon(uint width, uint height, Span<byte> pixels)
         {
             unsafe
             {
-                fixed (byte* PixelsPtr = pixels)
+                fixed (byte* ptr = pixels)
```
DemoXinMC, line 191: I'm personally fine with the changes to `ptr` vs `PixelPtr` everywhere, but that is purely semantics and it might be beneficial to actually create some coding style rules for the project that we don't currently have and this would be relevant to that discussion.

src/SFML.Graphics/RenderWindow.cs
```diff
@@ -565,44 +565,15 @@ public void Draw(Vertex[] vertices, PrimitiveType type)
...
+        public void Draw(Span<Vertex> vertices, PrimitiveType type, RenderStates states)
         {
             RenderStates.MarshalData marshaledStates = states.Marshal();
 
             unsafe
             {
                 fixed (Vertex* vertexPtr = vertices)
                 {
-                    sfRenderWindow_drawPrimitives(CPointer, vertexPtr + start, (UIntPtr)count, type, ref marshaledStates);
+                    sfRenderWindow_drawPrimitives(CPointer, vertexPtr, (UIntPtr)vertices.Length, type, ref marshaledStates);
```
DemoXinMC, line 576: Same as `RenderTexture`

src/SFML.Window/Clipboard.cs
```diff
@@ -29,10 +29,10 @@ public static string Contents
                     }
                 }
                 
-                byte[] sourceBytes = new byte[length * 4];
-                Marshal.Copy(source, sourceBytes, 0, sourceBytes.Length);
-                
-                return Encoding.UTF32.GetString(sourceBytes);
+                unsafe
+                {
+                    return Encoding.UTF32.GetString((byte*)source, (int)( length * 4 ));
```
DemoXinMC, line 34: While I know that UTF32 is going to yield 4 bytes per char, does it make more sense to use `sizeof()` for clarity on this and the other uses of `UTF*.GetString()`?

src/SFML.Graphics/Image.cs
```diff
@@ -310,14 +309,17 @@ public void SetPixel(uint x, uint y, Color color)
         /// </summary>
         /// <returns>Array of pixels</returns>
         ////////////////////////////////////////////////////////////
-        public byte[] Pixels
+        public Span<byte> Pixels
```
DemoXinMC, line 312: Is it reasonable to add a `set` method for `Pixels`?

src/SFML.Audio/SoundBufferRecorder.cs
```diff
@@ -57,9 +58,14 @@ protected override bool OnStart()
...
         /// <returns>False to stop recording audio data, true to continue</returns>
         ////////////////////////////////////////////////////////////
-        protected override bool OnProcessSamples(short[] samples)
+        protected override bool OnProcessSamples(Span<short> samples)
         {
-            mySamplesArray.AddRange(samples);
+            if (mySamplesArray.Capacity < mySamplesArray.Count + samples.Length)
+                mySamplesArray.Capacity = mySamplesArray.Count + samples.Length;
+
+            for (int i = 0; i < samples.Length; i++)
```
DemoXinMC, line 66: Is there not a way that we can use `AddRange()` with `Span`?  Also, Audio is likely going to benefit from a significant rework, but that's beyond the scope of this draft.

src/SFML.Audio/SoundBuffer.cs
```diff
@@ -181,13 +180,17 @@ public Time Duration
...
             {
-                short[] SamplesArray = new short[sfSoundBuffer_getSampleCount(CPointer)];
-                Marshal.Copy(sfSoundBuffer_getSamples(CPointer), SamplesArray, 0, SamplesArray.Length);
-                return SamplesArray;
+                ulong count = sfSoundBuffer_getSampleCount(CPointer);
+                IntPtr ptr = sfSoundBuffer_getSamples(CPointer);
+
+                unsafe
+                {
+                    return new Span<short>((void*)ptr, (int)count);
```
DemoXinMC, line 192: Since we .NET doesn't own the memory for samples, should we be handling this this way?  Our `Span` could be invalidated if CSFML moves things IIRC

src/SFML.Graphics/RenderTexture.cs
Marioalexsan: The way I see it, the second overload is just a span-like version of the first one, so users could just do `vertices.Slice(0, 3)` if they want to do offsets and lengths.

src/SFML.Audio/SoundBufferRecorder.cs
Marioalexsan: There is no such overload for `AddRange()` if I remember correctly.

charleyah:
Was there any consideration to use `ReadOnlySpan<T>` for the cases where SFML functions are reading from a span argument and not writing back to a caller provided buffer? At a glance that looks like the majority of cases.

Properties like `Image.Pixels` and `SoundBuffer.Samples` still make sense as `Span<T>` to allow callers to modify underlying data.

src/SFML.Audio/SoundBuffer.cs
charleyah: This will be safe for the lifetime of the SoundBuffer object instance for as long as all public members on SoundBuffer class don't produce iterator invalidation on the underlying `std::vector` containing the samples on `sf::SoundBuffer` (m_samples).

Marioalexsan:
@charleyah I haven't given it much thought initially, but yeah, we could use read-only spans in those places I think. I'd prefer to redo this PR after implementing all of the SFML.Net 3 changes though.

PR #269, opened 2024-08-17
Fix incorrect libraries for some DLLImport

Marioalexsan:
Seems like some DLLImports with incorrect library names slipped through the old PRs.


eXpl0it3r (approved):

eXpl0it3r:
Thanks for fixing this!

PR #270, opened 2024-08-25
Configure Stylecop rules

Marioalexsan:
# Description

(This is going to be a mess.)

This PR switches code style setup from `.editorconfig` + `omnisharp.json` to `.editorconfig` + `StyleCop.Analyzers` NuGet package. The main idea is to setup proper style rules that can also be enforced at compile time if needed.

I've supressed some of the StyleCop rules that didn't really fit SFML.Net's style (or my biases), while keeping others intact for now. No code changes are done yet, since 1) some changes that StyleCop wants would be breaking changes as far as the API is concerned, and 2) we might want to tweak it further until we've reached a comfortable setup.

# How to test?

Grab the branch, check the warnings generated by Intellisense, and check the warnings generated at compile time by the analyzer. Ideally also look through the code to check for other style changes that are needed.


Marioalexsan:
Closing this in favor of #271.

PR #277, opened 2024-09-20
SFML.Net 3

Marioalexsan:
This is nowhere near ready (CSFML v3 isn't even out), but I'm making this PR to track progress and to have higher visibility on the code changes.

The changes involved are as follows:
- remove obsolete code
- change methods that used multiple parameters for vector components (i.e. `int x, int y, int z`) to use vectors instead
- change P/Invokes to match the new (or expected) signatures on the CSFML side
- rename some members to match the C++ side (for example, `SystemHandle` -> `NativeHandle`)
- implement miniaudio changes for the Audio module
- extract  `Count` style enum values into a readonly field
- implement `Angle` and the new methods on `Vector2` and `Vector3`, use `Angle` instead of degrees expessed as floats
- implement `GetGeometricCenter` methods on shapes
- implement stencil and scissor features for Graphics module
- implement new methods for starting / stopping `Clock`
- added raw mouse event for Window
- added timeout parameter to `WaitEvent` style calls
- create basic unit tests (currently only `Angle` is tested)


src/SFML.Audio/Music.cs
```diff
@@ -165,8 +176,8 @@ public Music(byte[] bytes) :
         ////////////////////////////////////////////////////////////
         public bool Loop
```
eXpl0it3r, line 177:
```suggestion
        public bool IsLooping
```

Marioalexsan: Changed it for all sound objects

Marioalexsan:
The current PR seems to run with local CSFML 3.0 binaries (i.e. all example programs seem to run without any immediate issues), so I think it's now possible for people to test it out if they want to (although I'd expect it to crash in a lot of other places).

Since CSFML and SFML.Net 3.0 nuget packages are not available yet, here's instructions on how to build local CSFML and SFML.Net nuget packages for Windows / Visual Studio 2022:

For CSFML 3.0:
- Clone CSFML master branch
- Open a Powershell terminal in `tools/nuget` and run `./build.win.ps1`
- run `dotnet nuget locals all --clear` to clear previous cached packages if you need to (i.e. if you update the package code but use the same version)
- in `tools/nuget`, run `dotnet pack -c Release`
- in Visual Studio 2022, in `Tools -> NuGet Package Manager -> Package Sources`, add a path to `CSFML/tools/nuget/CSFML/bin/Release`

For SFML.Net 3.0:
- run `dotnet build -c Debug`
- in Visual Studio 2022, in `Tools -> NuGet Package Manager -> Package Sources`, add the following paths:
  - `SFML.Net\src\SFML.Net\bin\Debug`
  - `SFML.Net\src\SFML.Graphics\bin\Debug`
  - `SFML.Net\src\SFML.Audio\bin\Debug`
  - `SFML.Net\src\SFML.Window\bin\Debug`
  - `SFML.Net\src\SFML.System\bin\Debug`
- create a new project, and add the following to your `.csproj`:
```
<ItemGroup>
    <PackageReference Include="SFML.Net" Version="3.0.0-alpha1" />
</ItemGroup>
```

I've tested this flow with a new console project that uses one of the examples as source code, but let me know if it breaks in some other cases.

eXpl0it3r:
I've taken the liberty to update your branch by splitting the one commit into two, extracting the test stuff into a dedicated commit.  
Also you can't really (in an easy way) multi-target tests while mixing it in the same solution. So I just limited it to .NET 6

PR #296, opened 2026-01-16
Switch examples to .NET 8

eXpl0it3r:
SFML is primarily focussing on .NET Standard 2.0 and/or "mondern" .NET as such it makes sense to have all examples on .NET 8 and one example demonstrating .NET Framework support


examples/netfx/readme.md
Marioalexsan: This still mentions .NET Core, should be switched to explain it's about .NET Framwork.

Marioalexsan (approved):

